PR #3, opened 2018-09-02
Using official search api

martincognite:
Use the official search api instead of the old prefix search.

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -209,7 +209,7 @@ export default class CogniteDatasource {
       return Promise.resolve([]);
     }
     return this.backendSrv.datasourceRequest({
-      url: this.url + `/cogniteapi/${this.project}/timeseries?q=${query}`,
+      url: this.url + `/cogniteapi/${this.project}/timeseries/search?query=${query}`,
       method: "GET",
     }).then((result: { data: TimeSeriesResponse }) =>
       result.data.data.items.map(timeSeriesResponseItem => (


src/datasource.ts
```diff
@@ -209,7 +209,7 @@ export default class CogniteDatasource {
       return Promise.resolve([]);
     }
     return this.backendSrv.datasourceRequest({
-      url: this.url + `/cogniteapi/${this.project}/timeseries?q=${query}`,
+      url: this.url + `/cogniteapi/${this.project}/timeseries/search?query=${query}`,
```
wjoel: Doesn't this only work for projects where search has been enabled? If so, I'd prefer to not use it until it's enabled everywhere, and always, otherwise this plugin won't be "plug and play" for everyone.

I think it's not enabled for the "playground" project, for example.

martincognite: Fair enough.

martincognite:
Reopening this as an issue tracker for initial feedback from Framo.

It might be worth it to check if search is enabled for the tenant and conditionally ask the search api. It's pretty hard to search if you only get prefix search on the `name` property. What do you think @wjoel ? If nothing else we can block this issue with automatic search deployment for new tenants (which I think is the plan anyway).

wjoel:
@martincognite Sorry for the late reply, didn't see your message. Sounds like a good idea! I've added a card to the Trello board. Would be nice if we had a good way to check that through some API calll, but maybe it's already available and I'm just not aware of it.

martincognite:
This should probably be merged now as search is present on all tenants (and automatically gets deployed on tenant creation). @wjoel

PR #10, opened 2018-10-29
[WIP] Added lint + event datasource.

martincognite:
This should've been 2 smaller PRs -- but it wasn't until it was too late.

1. Quite intrusive PR, changes a lot of the code -- but only touches code-style rules (semicolons etc). Proposes `prettier` as a code standard and adds pre-commit hooks for in-place writing fixes
2. Adds support for `annotationQuery` with a very thin input layer, only `type` and `subType`. In addition.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,4 @@
+{
+  "trailingComma": "es5",
+  "singleQuote": true
+}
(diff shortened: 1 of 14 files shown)


martincognite:
@hulien22 this is not needed anymore right, can we close this?

hulien22:
Closing, implemented through #23 and #26

PR #11, opened 2018-10-29
Added Jenkinsfile for baking and pushing images to dockerhub

martincognite:


Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+          sh("docker build -t ${imageName}:${shortSha} .")
+        }
+
+        stage('Login to docker hub') {
+          sh('docker login -u "$(cat /dockerhub-credentials/DOCKER_USERNAME)" -p "$(cat /dockerhub-credentials/DOCKER_PASSWORD)"')
+        }
+
+        if (env.CHANGE_ID) {
+          stage("Publish PR image") {
+            def prImage = "${devImageName}:pr-${env.CHANGE_ID}"
```
wjoel: I don't think this makes sense for this repository?

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+  ) {
+    node(label) {
+      def shortSha
+      container('node') {
+        stage('Checkout') {
+          checkout(scm)
+          shortSha = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
+        }
+
+        stage('Prepare') {
```
wjoel: Build and test, or keep `'Prepare'` but then it should just do `yarn` and not `yarn build` (which would then be in a `'Build'` stage)

Jenkinsfile
martincognite: The plan was to ship this to a `grafana-cdp-dev` if someone wanted to run the artifact without cloning, building and running locally.

Jenkinsfile
martincognite: Good point.

Jenkinsfile
```diff
@@ -0,0 +1,81 @@
+def label = "cognite-grafana-datasource-${UUID.randomUUID().toString().substring(0, 5)}"
+def imageName = "cognite/grafana-cdp"
+def devImageName = "cognite/grafana-cdp-dev"
+
+podTemplate(
+  label: label,
+  containers: [
+    containerTemplate(
+      name: 'node',
+      image: 'node:9',
```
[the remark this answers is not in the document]
wjoel: No need to be frugal here either, `1000m` or even `2000m` CPU is fine.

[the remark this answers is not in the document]
martincognite: Done done.

Jenkinsfile
```diff
@@ -0,0 +1,83 @@
...
+  envVars: [
+    envVar(key: 'CHANGE_ID', value: env.CHANGE_ID),
+  ],
+  ) {
+    node(label) {
+      def shortSha
+      container('node') {
+        stage('Checkout') {
+          checkout(scm)
+          shortSha = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
```
wjoel: We've found it quite useful to include the date in the image tag, see https://github.com/cognitedata/transformers-backend/blob/master/Jenkinsfile#L48 for an example.

Issue #12, opened 2018-11-23
Use "aggregates" rather than "aggregation" when calling /dataquery


larshagencognite:

hulien22:
Fixed in #21 (specifically [71c8b18](https://github.com/cognitedata/cognite-grafana-datasource/commit/71c8b182899871330754bea88a7730f66adcbe49)). Closing now.

PR #13, opened 2018-11-23
Remove incorrect dataquery parameter

wjoel:
Aggregates are already specified per time series in the query items, so this parameter should not be necessary.

The time series team wants to enforce more strict query parameter checking, so we need to remove this.


wjoel:
Should fix https://github.com/cognitedata/cognite-grafana-datasource/issues/12

@johhat do you have time to test this locally?

wjoel:
This isn't quite this simple, retrieving the data works fine but then Grafana isn't showing it when using aggregates. Probably because it's looking for the wrong field name with this change.

@johhat will look into it when he gets the chance.

wjoel:
Replaced by #14

PR #21, opened 2019-01-23
Asset integration

hulien22:
- Allowing for variables to be used in the `Root Asset` field
- Small fixes

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,9 @@
     "typescript": "^3.0.1"
   },
   "dependencies": {
+    "@types/react": "^16.7.21",
     "lodash": "^4.17.10",
-    "q": "^1.5.1"
+    "q": "^1.5.1",
+    "react": "^16.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -10,10 +10,10 @@
     "run-docker": "docker build -t cognite-grafana . && docker run -p 3000:3000 cognite-grafana"
   },
   "devDependencies": {
+    "@types/react": "^16.7.22",
     "typescript": "^3.0.1"
   },
   "dependencies": {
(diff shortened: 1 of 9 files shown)


src/datasource.ts
```diff
@@ -275,7 +289,30 @@ export default class CogniteDatasource {
           aggregation: q.aggregates,
         }
       }));
-    }, []);
+
+      // assign labels to each timeseries
+      if (target.tab === Tab.Timeseries) {
+        if (target.label.indexOf("{{") < target.label.lastIndexOf("}}")) {
```
martincognite: This is beautiful, but it looks to me like it will match with the label expression `labels are cool}}`, as (`indexOf("{{")` would be -1 which is < whatever index `}}` is on.

hulien22: Good point, I'll change it to do a simple regex match on /{{.*}}/

martincognite:
`Panel data error: TypeError: Cannot read property 'timeseries' of undefined`
This keeps popping up in the console on new dashboards

hulien22:
> `Panel data error: TypeError: Cannot read property 'timeseries' of undefined`
> This keeps popping up in the console on new dashboards

This is fixed now.

package.json
```diff
@@ -13,7 +13,9 @@
     "typescript": "^3.0.1"
   },
   "dependencies": {
+    "@types/react": "^16.7.21",
```
martincognite: nit, this should be a `devDependency` (`yarn add @types/react --dev`)

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -p 3000:3000 cognite/grafana-cdp-dev:pr-21`

PR #27, opened 2019-02-06
Improved functions

hulien22:
Allowing for aggregations to be used within functions, via the syntax [ID,aggregation,granularity]
For example: 
```  
timeseries{name=~'.*temp.*', function = ([ID] - 32) * 5/9}
timeseries{ function = [ID] - [ID,avg,24h] }
timeseries{ function = [ID,step] - [123456789,step] }
```

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -125,6 +125,13 @@ export interface DataDatapoints {
   data: Datapoints;
 }
 
+interface DataQueryAlias {
+  alias: string;
+  id: number;
+  aggregate?: string;
+  granularity?: string;
+}
+
 interface DataQueryRequestItem {
   name: string;
   start?: string | number;
@@ -133,6 +140,7 @@ interface DataQueryRequestItem {
   granularity?: string;
   aggregates?: string;
   function?: string;
+  aliases?: DataQueryAlias[];
 }
 
 interface DataQueryRequest {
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -p 3000:3000 cognite/grafana-cdp-dev:pr-27`

src/datasource.ts
```diff
@@ -315,6 +323,26 @@ export default class CogniteDatasource {
...
         if (target.tab === Tab.Custom && target.assetQuery.func) {
           query.function = target.assetQuery.func.replace(/ID/g, String(ts.id));
+          const regexSearch = /\[.*?\]/g;
+          const regexMatches = query.function.match(regexSearch);
+          query.aliases = [];
+          for (const match of regexMatches) {
+            // format is id, aggregation, granularity
+            const aliasParts = match
+              .substr(1, match.length - 2)
+              .split(',')
```
martolini: Keep in mind that `','.split(',')` yields `['', '']`. Might want to do an `.filter(string => string.length)` or similar.

src/datasource.ts
```diff
@@ -315,6 +323,26 @@ export default class CogniteDatasource {
...
+              .map(x => _.trim(x, ' \'"'));
+            if (aliasParts.length === 1) continue;
```
martolini: Why aren't we happy with a length of one? Is this a way of saying that there's no commas (i.e only `ID` is specified)?

src/datasource.ts
```diff
@@ -315,6 +323,26 @@ export default class CogniteDatasource {
...
+            const alias: DataQueryAlias = {
+              alias: `alias${aliasParts.join('_')}`,
+              id: Number(aliasParts[0]),
+            };
+            alias.aggregate = aliasParts[1];
```
martolini: This is out of scope for this PR, but it might be nice to make an `Enum` with the aggregates as some point as they are static.

src/datasource.ts
```diff
@@ -390,7 +418,11 @@ export default class CogniteDatasource {
           ids += Object.keys(idsObj).length;
         }
         if (ids === 0) ids = 1; // will fail anyways, just show the api error message
-        queryReq.limit = Math.floor(100_000 / ids);
+
+        // check if any aggregates are being used
+        const usesAggregations = queryList.some(item => item.aliases.length > 0);
+
+        queryReq.limit = Math.floor((usesAggregations ? 10_000 : 100_000) / ids);
```
martolini: This `Math.floor` calculations is starting to appear multiple places now, we can consider creating a `calculateLimit` function or something.

src/datasource.ts
```diff
@@ -689,7 +721,8 @@ export default class CogniteDatasource {
 
     let splitfilters: string[];
     if (timeseriesMatch) {
-      splitfilters = timeseriesMatch[1].split(',');
+      // regex finds commas that are not followed by a closed bracket
+      splitfilters = _.split(timeseriesMatch[1], /,(?![^\(\[]*[\]\)])/g);
```
martolini: Holy moly.

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -p 3000:3000 cognite/grafana-cdp-dev:pr-27`

src/datasource.ts
hulien22: If we only get `[ID]` then it's better to not make an alias, as if we don't have any aliases then we can use a max limit of 100000. As soon as we make an alias, this limit goes down to 10000, so this is just a small optimization. 

Added a comment to the code to explain this.

src/datasource.ts
hulien22: fixed

src/datasource.ts
hulien22: This only shows up in two places, and are right next to each other - no need right now to make such a function.

PR #30, opened 2019-02-18
Optimizing requests

hulien22:
Adding a cache so that we don't send redundant requests. 
Also changing the dataquery requests to be sent in parallel instead of sequentially

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,0 +1,53 @@
+import { isError } from './datasource';
+
+// Cache requests for 10 seconds
+const cacheTime = 1000 * 10;
+
+const queries = {
+  results: new Map(),
+  requests: new Map(),
+};
+
+export const getQuery = async (query, backendSrv) => {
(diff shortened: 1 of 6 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -p 3000:3000 cognite/grafana-cdp-dev:pr-30`

src/cache.ts
```diff
@@ -0,0 +1,53 @@
...
+        // the item may not exist, or it may have been deleted
+        return {};
+      }
+      const asset = res;
+      if (!isError(asset)) {
+        queries.results.set(stringQuery, asset);
+        // set a timeout to clear the cache
+        setTimeout(() => {
+          queries.results.delete(stringQuery);
+          queries.requests.delete(stringQuery);
```
martincognite: Deleting the request shouldn't be necessary (nor harmful for that matter).

PR #43, opened 2019-03-06
V0.0.1 feedback fixes

hulien22:
Fixing and changing a few things based off of feedback given by Grafana. 
- Fixes the checkboxes in Select Timeseries from Asset, and adds a select all option
- Changes the pattern used for inputting api-keys, where once the api-key is entered, you must press a reset button to reset it; Also, changed it to test for the project name when a user hits Save and Test.

### DIFF
--- a/src/css/query_editor.css
+++ b/src/css/query_editor.css
@@ -33,5 +33,5 @@ input[type='checkbox'] {
 }
 
 .cognite-timeseries-list-checkbox {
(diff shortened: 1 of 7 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -p 3000:3000 cognite/grafana-cdp-dev:pr-43`

Issue #36, opened 2019-02-26
Move timeseries out of the queryTarget object


hulien22:
We currently store all timeseries rooted at an asset in the `QueryTarget` object. This was fine as that object was only really visible through the JSON section. However, it really should not be there as that object should just be for properties that remain constant between reloads of the page (ie not for storage)

Grafana 6.0.0 also introduces the explore section, which puts the querytarget object in the url, and this can cause issues if we try to store a lot of timeseries.

hulien22:
I forgot that we do actually require the timeseries array for 'Select Timeseries from Asset' so we can't completely remove it...
We may want to look into only storing whether or not a timeseries is selected in the `QueryTarget` object, that way we can still reduce its size.

hulien22:
We now only keep the timeseries array for 'Select Timeseries from Asset' and no longer in 'Custom Query', and also limit the size of that array. This has been fixed to the point that explore is usable and the QueryTarget object is smaller, so I'm closing this issue.

PR #44, opened 2019-03-11
Optimizing QueryTarget Object

hulien22:
Since we have no need to save all asset timeseries for custom queries, I am adding a cache to store the timeseries separately from the `QueryTarget` object. We still need to keep the timeseries array in `QueryTarget` since `Get Timeseries from Asset` still needs to keep track of which ones are selected. However, when we load in the target object in `query_ctrl` we now also check if we are on the assets tab, if not we can clear that array.

Also choosing to limit `Get Timeseries from Asset` to only fetch the first 100 timeseries since `/dataquery` can only show 100 items, and also because that section really should not be used for that many timeseries. A warning will be shown if not all timeseries can be shown. 

Mostly solves #36, although we may be able to further optimize `QueryTarget` by reducing how much we store for the each timeseries in the timeseries array. Also, with the limitation to 100 timeseries, explore no longer has issues with the url length (at least in chrome).

As an extra note, this will help when we have a way to concurrently fetch all timeseries belonging to an asset, as we won't have to store them in `QueryTarget`.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -6,6 +6,7 @@ import {
   QueryOptions,
 } from './types';
 import { BackendSrv } from 'grafana/app/core/services/backend_srv';
+import Utils from './utils';
 
 // Cache requests for 10 seconds
 const cacheTime = 1000 * 10;
@@ -56,10 +57,7 @@ export const getQuery = async (query: DataSourceRequestOptions, backendSrv: Back
 const assetTimeseries = new Map<string, TimeSeriesResponseItem[]>();
 
 export const getTimeseries = (options: QueryOptions, target: QueryTarget) => {
-  const id: string = `${options.dashboardId}_${options.panelId}_${target.refId}_${
-    target.assetQuery.templatedTarget
-  }_${target.assetQuery.includeSubtrees}`;
-  return assetTimeseries.get(id);
+  return assetTimeseries.get(Utils.timeseriesHash(options, target));
 };
 
 export const setTimeseries = (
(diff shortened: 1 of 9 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-44`

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-44`

src/cache.ts
```diff
@@ -47,8 +53,30 @@ export const getQuery = async (query: DataSourceRequestOptions, backendSrv: Back
   return promise;
 };
 
+const assetTimeseries = new Map<string, TimeSeriesResponseItem[]>();
+
+export const getTimeseries = (options: QueryOptions, target: QueryTarget) => {
+  const id: string = `${options.dashboardId}_${options.panelId}_${target.refId}_${
```
martincognite: a hashTimeseries object or similiar would be nice, this looks prone to error (and is used multiple places) ;-)

src/datasource.ts
```diff
@@ -414,13 +456,15 @@ export default class CogniteDatasource {
...
 
-    const searchQuery: Partial<TimeseriesSearchQuery> = {
-      path: target.assetQuery.includeSubtrees ? [assetId] : undefined,
-      assetId: !target.assetQuery.includeSubtrees ? assetId : undefined,
-      limit: 10000,
-    };
-
+    // since /dataquery can only have 100 items and checkboxes become difficult to use past 100 items,
+    //  we only get the first 100 timeseries, and show a warning if there are too many timeseries
+    searchQuery.limit = 101;
```
martincognite: This is very clever from a UI/UX perspective.

src/cache.ts
hulien22: added a timeseriesHash() function

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-44`

PR #45, opened 2019-03-12
Fixing dropdown display

hulien22:
Forcing update once we get options for a dropdown, since otherwise it doesn't always get updated and is just empty.


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-45`

PR #49, opened 2019-03-25
allowing custom queries to return more than 100 timeseries

### LINKED ISSUE #48: Allow custom queries to produce more than 100 timeseries
Currently, we get an error saying that the limit for items is 100, which is not very useful to a user as they have little control over this. 
We should be able to create additional `/dataquery` calls with a max of 100 each

hulien22:
There is currently a limit for `/dataquery` to only contain 100 items per request, but this should not limit a user
Now checking if a request will contain more than 100 items and if it does, split it up into more requests.
Fixes #48

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -168,7 +168,7 @@ export default class CogniteDatasource {
     const dataQueryRequestItems = await Promise.all(dataQueryRequestPromises);
 
     const queries: DataQueryRequest[] = [];
-    for (let { target, queryList } of dataQueryRequestItems.map((ql, i) => ({
+    for (const { target, queryList } of dataQueryRequestItems.map((ql, i) => ({
       target: queryTargets[i],
       queryList: ql,
     }))) {
@@ -177,17 +177,17 @@ export default class CogniteDatasource {
       }
 
       // /dataquery is limited to 100 items, so we need to add new calls if we go over 100 items
-      while (queryList.length > 0) {
-        const slicedQl = queryList.slice(0, 100); // only get first 100 items
-
+      // may want to change how much we split into, but for now, 100 seems like the best
+      const qlChunks = _.chunk(queryList, 100);
+      for (const qlChunk of qlChunks) {
         // keep track of target lengths so we can assign errors later
         targetQueriesCount.push({
           refId: target.refId,
-          count: slicedQl.length,
+          count: qlChunk.length,
         });
         // create query requests
         const queryReq: DataQueryRequest = {
-          items: slicedQl,
+          items: qlChunk,
           start: timeFrom,
           end: timeTo,
         };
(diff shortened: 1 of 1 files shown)


martincognite:
Haven't reviewed yet, but I've heard that 10 is the optimal number of time series to ask for in dataquery. If we're shipping requests in parallell anyway I mean.

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-49`

src/datasource.ts
```diff
@@ -168,61 +168,70 @@ export default class CogniteDatasource {
...
-        const idRegex = /\[.*?\]/g; // look for [something]
-        for (const q of queryList) {
-          const matches = q.function.match(idRegex);
-          if (!matches) break;
-          const idsObj = {};
-          for (const match of matches) {
-            idsObj[match.substr(1, match.length - 2)] = true;
+
+      // /dataquery is limited to 100 items, so we need to add new calls if we go over 100 items
+      while (queryList.length > 0) {
```
martincognite: lodash has a chunk function, which will split the querylist into objects of `n`. Might be easier to debug than to constantly mutate the running object.

src/datasource.ts
```diff
@@ -168,61 +168,70 @@ export default class CogniteDatasource {
...
+          if (queryReq.aggregates) {
+            target.error =
+              '[ERROR] To use aggregations with functions, use [ID,aggregation,granularity] or [ID,aggregation]';
+            break;
+          }
+          let ids = 0;
+          const idRegex = /\[.*?\]/g; // look for [something]
+          for (const q of slicedQl) {
+            const matches = q.function.match(idRegex);
+            if (!matches) break;
```
martincognite: Do you intentionally break here? Or do you want to `continue` to the next `q`.

src/datasource.ts
```diff
@@ -168,61 +168,70 @@ export default class CogniteDatasource {
...
-          ids += Object.keys(idsObj).length;
         }
-        if (ids === 0) ids = 1; // will fail anyways, just show the api error message
+        if (target.assetQuery && target.assetQuery.func && target.tab === Tab.Custom) {
+          if (queryReq.aggregates) {
+            target.error =
+              '[ERROR] To use aggregations with functions, use [ID,aggregation,granularity] or [ID,aggregation]';
+            break;
+          }
+          let ids = 0;
```
martincognite: Would be nice to know that this isn't actually `ids`, it's a `idsCount` or similarly. Becomes extra tricky further down, with `/ slicedQl.length` and `/ ids` (and not `/ ids.length`)

src/datasource.ts
hulien22: Each of these queries all belong to the same query so they would all have the same function. So if one of them doesn't have an id in its function, then none of them do

src/datasource.ts
hulien22: makes sense, will change

src/datasource.ts
hulien22: yeah that does sound simpler, I'll fix that

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-49`

hulien22:
I tried splitting into chunks of 10, 50 and 100 - still seems like 100 had the best performance (probably because we are being limited by the number of calls we can send in parallel?) 
Something to note is that since limits affect the entire request, having smaller chunk sizes could result in more datapoints for some requests.

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-49`

PR #50, opened 2019-03-26
fixing issue with labels for custom queries

hulien22:
now if no label is specified for a custom query, we will just use the name of the last timeseries in the function


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-50`

PR #53, opened 2019-03-28
Adding warning when we hit the datapoint limit

hulien22:
Adding this now so that we at least have something when we hit the datapoints limit

We may decide to improve on this later - for example, by repeating `/dataquery` calls until we get all the data in

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -221,15 +221,21 @@ export default class CogniteDatasource {
         const aggregation = response.config.data.aggregates;
         const aggregationPrefix = aggregation ? `${aggregation} ` : '';
         return datapoints.concat(
-          response.data.data.items.map(item => ({
-            target: labels[count++] ? labels[count - 1] : aggregationPrefix + item.name,
-            datapoints: item.datapoints
-              .filter(d => d.timestamp >= timeFrom && d.timestamp <= timeTo)
-              .map(d => {
-                const val = Utils.getDatasourceValueString(response.config.data.aggregates);
-                return [d[val] === undefined ? d.value : d[val], d.timestamp];
-              }),
-          }))
+          response.data.data.items.map(item => {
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-53`

src/datasource.ts
```diff
@@ -221,15 +221,21 @@ export default class CogniteDatasource {
...
-            target: labels[count++] ? labels[count - 1] : aggregationPrefix + item.name,
-            datapoints: item.datapoints
-              .filter(d => d.timestamp >= timeFrom && d.timestamp <= timeTo)
-              .map(d => {
-                const val = Utils.getDatasourceValueString(response.config.data.aggregates);
-                return [d[val] === undefined ? d.value : d[val], d.timestamp];
-              }),
-          }))
+          response.data.data.items.map(item => {
+            if (item.datapoints.length >= response.config.data.limit) {
```
martincognite: Is this per request? Don't we have to sum of the count of all the limits instead of each individual ts?

hulien22: This checks if any of the timeseries returned from `/dataquery` hits its limit.

For example, if you send a request for ten timeseries with aggregates, then each timeseries gets a limit of 10000/10 = 1000. Then if any of these ten timeseries return 1000 datapoints, then they probably didn't return all the datapoints for that timerange - so we would show this warning

hulien22: The whole point of this warning is just to warn users that they may not have gotten all the datapoints for their selected time range and to remove any confusion as to why their timeseries just ends abruptly in the middle

Again, this is just a quick and easy solution to this issue, we can improve on this later by pulling in all the datapoints for a time range - but that currently has a bunch of issues with limits and knowing where to start the next `/dataquery` call from

PR #55, opened 2019-04-04
Replace variables outside filters

hulien22:
Adding more variable support. 
`$Variable` or `[[Variable]]`  can now be used in labels, as well as outside of `timeseries{}` filters in custom queries (this allows you to do things like make a custom text box variable, and then having `timeseries{} $Variable`, thus making the query completely customizable via variables (set `$Variable` to things like ` + 100` or `* 3600`, etc.))


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-55`

PR #67, opened 2019-05-18
Allowing for the base url to be configured

### LINKED ISSUE #66: Make base url configurable
The base url should be configurable so that the data source can be used for projects in greenfield and when running the api locally.

hulien22:
Closes #66.

A little tricky because of backwards compatibility issues, since we are introducing a new template variable to the routes (which need a server reload upon any changes), which means that I got to learn about how Go's `template` works.
For any future references, [this is the code Grafana uses to update parts of the url](https://github.com/briangann/grafana/blob/d64971e35b09e13ff95624cadbe11141e2ae27eb/pkg/api/pluginproxy/utils.go#L30-L49).

### DIFF
--- a/src/partials/config.html
+++ b/src/partials/config.html
@@ -45,4 +45,23 @@
       </a>
     </div>
   </div>
+  <div class="gf-form-inline">
+    <div class="gf-form">
+      <span class="gf-form-label width-9">API url</span>
+      <input
+        type="text"
+        class="gf-form-input"
+        ng-model="ctrl.current.jsonData.cogniteApiUrl"
+        placeholder="api.cognitedata.com"
+      >
+      <info-popover mode="right-absolute">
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-67`

hulien22:
Wasn't actually able to test this with a non-`api.cognitedata.com` project, so if @erlendvollset or @martincognite could test that for me that'd be great :)

hulien22:
hmm, that's pretty weird since if the request went through and we got an empty response, that probably means there was an issue outside the connector - since aggregates still work in `api.cognitedata.com` and none of that code was touched

martincognite:
Agreed, so LGTM!

Issue #63, opened 2019-05-14
Implement Retrying


hulien22:
It'd be nice to have a way for the data source to automatically retry some failed requests. Would help to mitigate some issues that arise from rate limiting. 
First thoughts are to just implement retrying 5 times with some sort of backoff.

hulien22:
Something to consider here is that if there is a short refreshing value or if the requests take a long time, we may start to have overlapping requests (in which a newer request finishes before an older request and then doesn't get shown). This is already a small issue currently, but does not show up very often.

This means having some sort of cancellation tokens, or keeping track of all requests per row and whether or not there have been newer requests.

polomani:
Rate limits are currently not an issue, especially after the introduction of OIDC tokens.
The ticket will be reopened in Jira if similar issues start happening again.

PR #68, opened 2019-05-27
Adding request cancellation

hulien22:
Working on #63, and was looking at https://github.com/grafana/grafana/blob/master/public/app/core/services/backend_srv.ts in order to see if Grafana has implemented any type of retrying - but it seems like the retry option that is there is only used for grafana 401 errors.

Either way, I was able to find out that `backendSrv` does support cancelling requests that have the same `requestId` if a newer request is made, so I added in that support. Currently only adding this for `dataquery` requests (using `dashboardId_panelId_targetRefId` as the `requestId`) as most other requests don't really need this functionality. 
This should help to make #63 a bit easier to implement now, since we don't have to worry about cancellation tokens (see my comments there)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -155,4 +155,9 @@ export default class Utils {
       target.assetQuery.templatedTarget
     }_${target.assetQuery.includeSubtrees}`;
   }
+
+  // used for generating the options.requestId
+  static getRequestId(options: QueryOptions, target: QueryTarget) {
+    return `${options.dashboardId}_${options.panelId}_${target.refId}`;
+  }
 }
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -17,7 +17,7 @@ const queries = {
 };
 
 export const getQuery = async (query: DataSourceRequestOptions, backendSrv: BackendSrv) => {
-  const stringQuery = JSON.stringify(query);
+  const stringQuery = JSON.stringify({ ...query, requestId: undefined });
 
   if (queries.requests.has(stringQuery)) {
     return queries.requests.get(stringQuery);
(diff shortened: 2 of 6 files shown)


cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-68`

src/utils.ts
```diff
@@ -155,4 +155,9 @@ export default class Utils {
       target.assetQuery.templatedTarget
     }_${target.assetQuery.includeSubtrees}`;
   }
+
+  // used for generating the options.requestId
+  static getRequestId(options: QueryOptions, target: QueryTarget) {
+    return `${options.dashboardId}_${options.panelId}_${target.refId}`;
```
martincognite: Don't we need to encode the time ranges and stuff per query?

hulien22: Not for this, we are cancelling requests that come from the same target (ie each row for a graph). I think we only really need `panelId` and `targetRefId`, but I'm including the `dashboardId` just in case Grafana adds a way in the future to show multiple dashboards at once. 

The point of this is that if I send a bunch of expensive queries, for example if I'm under `Select Timeseries from Asset` and I'm clicking a bunch of checkboxes, each time I click one I'll be sending a new request and canceling the previous request. 
_(Currently, there is the potential for this: The user selects the `Select All` checkbox sending Request A, and then deselects a checkbox sending Request B immediately after. Now let's suppose Request B comes back first, Grafana updates the graph to go with Request B data. Then once Request A resolves and returns, Grafana will now update with Request A data. So we now have a graph that does not correspond with the query.)_

hulien22: But I did realize from your comment that I need to make a small change to the cache, since if I'm now encoding the requestId, we'll never use the results from a request for a duplicate target

cognite-cicd:
[pr-bot]
Run this build with `docker run --rm -it -p 3000:3000 cognite/grafana-cdp-dev:pr-68`

PR #74, opened 2019-08-05
V1 update

hulien22:
Converting api calls from v0.5 to v1.
- Currently, all api calls except for `/dataquery` have been converted to use the corresponding v1 endpoints (that still requires v0.5 synthetic timeseries). 
- Additionally, we are no longer using `assetSubtrees` or similar calls - this has been replaced with cursoring + paging through child assets (something to note is that we are still using the old limits here, as going above severely affects performance)

Things left to do:
- need to determine if we still want to support `assetSubtree` for events/annotations and assets/variables. I believe this is being used by people, and removing that support would affect backwards compatibility (it's very useful when used with template variables, eg `event{assetSubtrees=[$VAL_WELLS]}`). 
    - Supporting this would require additional logic to check if that filter is being used, and then constructing additional calls by fetching the subassets and then use those as part of the `parentId` filter parameter. 
    - Another option would be to remove that support, and instead implement cursoring here (currently isn't implemented). Again, not sure what effect this might have on performance, but I have had issues in the past with Grafana crashing when there are a ridiculous number of events being shown (something that would be a larger possibility if we allow cursoring)
- assets/search doesn't support query, so we need to do some parallel calls for the asset search dropdown to work as before
- determining what we want to do with limits?
- fixing tests + a lot more tests to make sure nothing is broken
- extra documentation for stuff that's been changed - especially relevant if we keep `assetSubtrees` in annotations/variables

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -8,15 +8,16 @@ import {
 import { BackendSrv } from 'grafana/app/core/services/backend_srv';
 import Utils from './utils';
 
-// Cache requests for 10 seconds
-const cacheTime = 1000 * 10;
-
 const queries = {
   results: new Map(),
   requests: new Map(),
 };
 
-export const getQuery = async (query: DataSourceRequestOptions, backendSrv: BackendSrv) => {
+export const getQuery = async (
+  query: DataSourceRequestOptions,
+  backendSrv: BackendSrv,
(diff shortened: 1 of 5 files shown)


hulien22:
fyi @martincognite

hhenriques:
Closed in favour of: https://github.com/cognitedata/cognite-grafana-datasource/pull/87

PR #84, opened 2019-11-25
Events as tables

hulien22:


hhenriques:
This is no longer aligned with the new version of the connector.

PR #86, opened 2019-12-17
fixing parser for special functions and regex groups

hulien22:
Fixes the case where we have something like:
`Sum(timeseries{name=~"(Gas|Oil).*"}[avg])`

Previously the parser would get confused by the `)` from the regex group filter.


hhenriques:
This'll be fixed with the v1 version.

PR #96, opened 2020-03-24
fix: parse annotation query with variable in it

mrtraser:


mrtraser:
> ok for now, maybe will come up with something else later

it's about making it work, guess we'll discuss how to deal with error handling later

PR #97, opened 2020-03-26
fix: synthetics rebased

polomani:


package.json
```diff
@@ -57,13 +57,14 @@
...
+      "tslint -c tslint.json --fix",
       "prettier --write",
       "git add"
     ],
     "*.{md}": [
       "prettier --write",
       "git add"
-    ]
+    ],
+    "ignore": ["*/**/grammar.ts"]
```
mrtraser: nice)

src/parser/ts/grammar.ne
```diff
@@ -0,0 +1,196 @@
...
+  | "null" {% () => null %}
+
+variable -> "$" prop_name {% extractVariable %}
+  | "[[" prop_name "]]" {% extractVariable %}
+
+operator -> _ "+" _ {% extractOperator %}
+  | _ "-" _ {% extractOperator %}
+  | _ "/" _ {% extractOperator %}
+  | _ "*" _ {% extractOperator %}
+comma -> _ "," _ {% d => d[1] %}
```
mrtraser: it's actually `{% extractOperator %}`, right?

src/parser/ts/grammar.ne
```diff
@@ -0,0 +1,196 @@
...
+  | "abs" {% id %}
+  | null {% id %}
+binary -> "pow" {% id %}
+  | "round" {% id %}
+  | "on_error" {% id %}
+n_ary -> "max" {% id %}
+  | "min" {% id %}
+  | "avg" {% id %}
+  | "sum" {% id %}
+br ->  _ "(" {% null %}
```
mrtraser: looks a bit weird to me, harder to read

src/parser/ts/index.ts
```diff
@@ -0,0 +1,525 @@
...
+const compiledGrammar = Grammar.fromCompiled(grammar);
+
+export const formQueriesForExpression = async (
+  expr: string,
+  options: QueryOptions,
+  templateSrv: TemplateSrv,
+  target: QueryTarget,
+  connector: Connector
+): Promise<DataQueryRequestItem[]> => {
+  const templatedExpr = templateSrv.replace(expr.trim(), options.scopedVars);
```
mrtraser: do you really need this service here? We can replace variables before passing expr here

src/parser/types.ts
```diff
@@ -1,12 +1,13 @@
...
+  RegexNotEquals: '!~' as FilterType,
+  RegexEquals: '=~' as FilterType,
+  NotEquals: '!=' as FilterType,
+  Equals: '=' as FilterType,
+};
+export type FilterType = '!~' | '=~' | '!=' | '=';
 export interface QueryFilter {
   filter: FilterType;
-  value: string;
+  value: string | number;
```
mrtraser: I thought that filter value could be only `string` that will be passed to the regexp. Why `number` here?

src/query_ctrl.ts
```diff
@@ -80,6 +81,16 @@ export class CogniteQueryCtrl extends QueryCtrl {
     });
   }
 
+  onChangeCustom() { // todo: why it doesn't work without this hack?
+    try {
+      parse(this.target.expr);
```
mrtraser: this will not work cause it doesn't trigger `$digest` circle to render the component. If we want to handle errors for STS, we should use `$scope` params, i.e. change `$scope` params

mrtraser: oh, actually it works. So what's the hack here?

src/cdfDatasource.ts
```diff
@@ -91,40 +97,27 @@ async function getLabelsForTarget(
...
-          if (useExternalId) {
-            labels.push(ts[i].externalId);
-          } else {
-            labels.push(getLabelWithInjectedProps(target.label || '', ts[i]));
-          }
+    case Tab.Timeseries: {
+      return [await getTimeseriesLabel(target.label, target.target, target, connector)];
+    }
+    case Tab.Asset: {
+      const labels = [];
```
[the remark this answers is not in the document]
polomani: hehe, but I didn't actually touch this part, maybe keep it for now?
just trying to limit the scope of pr

src/cdfDatasource.ts
```diff
@@ -33,30 +33,36 @@ export function formQueryForItems(
...
-      aggregates: [aggregation],
-      granularity: granularity || ms2String(options.intervalMs),
+  if (tab === Tab.Custom) {
+    const hasAggregates = false; // todo: check if there are ny aggregations inside
+    const limit = calculateDPLimitPerQuery(hasAggregates, items.length);
+    return {
+      items: items.map(({ expression }) => ({ expression, start, end, limit })),
+    };
+  }
+  {
```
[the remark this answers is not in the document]
polomani: fixed, pushed rn

src/parser/ts/grammar.ne
polomani: comma? no, I just drop it later. operator is for things like +/-*, I don't need comma in the end result

src/cdfDatasource.ts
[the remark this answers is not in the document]
mrtraser: yep, it was the thing that comes from replacing `labels` for `Tab.Timeseries` case.

src/parser/ts/grammar.ne
mrtraser: got it

src/parser/ts/index.ts
polomani: good point, fixed now

src/parser/events-assets/index.ts
```diff
@@ -4,14 +4,14 @@ import getPaths from 'deepdash/getPaths';
...
+import { ParserResponse, QueryParserResponse } from '../types';
 
+const compiledGrammar = Grammar.fromCompiled(grammar);
 const filterDeep = getFilterDeep(_);
 const paths = getPaths(_);
 const omitDeep = getOmitDeep(_);
 
-const parseQuery = (query: string): QueryParserResponse => {
-  const parser = new Parser(Grammar.fromCompiled(grammar));
+export const parseWith = (parser: Parser, query: string) => {
```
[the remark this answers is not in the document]
polomani: fixed

src/parser/ts/grammar.ne
polomani: the reason I have is that I don't need to have a bunch of space parameters in extractor (which is even worse to read).

Maybe I could rename it or something?

src/cdfDatasource.ts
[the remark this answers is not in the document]
polomani: will change it to
```
const labelSrc = target.label || '';
      return target.assetQuery.timeseries
        .filter(ts => ts.selected)
        .map(ts => getLabelWithInjectedProps(labelSrc, ts))
```
after all 😛

src/parser/ts/grammar.ne
mrtraser: I see your point. No, don't think you'll find a better name, let's leave it as is

src/query_ctrl.ts
polomani: the things is that we have error handling deeper in the code, like in `handleFailedTargets`, but when I try assign errors to target by doing `target.error='message'` it doesn't work. And that's actually preferred way for me, since then we we can show all other errors, like http. but it doesn't display the error for some reason.
I though maybe you could find the reason (it's very angular specific for me)

src/parser/types.ts
polomani: good point, fixed that now. changed types slightly as well

src/query_ctrl.ts
mrtraser: Here is a thing, that we can't control rendering from the `query` method. What we can do here is to use `appEvents` grafana event emitter. I'll come up with a solution tmrw.

src/__tests__/utils.ts
```diff
@@ -47,8 +48,10 @@ const getTemplateSrvMock = () =>
     replace: jest.fn((q, options) => {
       let query = q;
       for (const { name, current } of variables) {
-        query = query.replace(`[[${name}]]`, current.value);
-        query = query.replace(`$${name}`, current.value);
+        while(query.includes(`[[${name}]]`) || query.includes(`$${name}`)) {
```
mrtraser: can we have proper regex here?
```typescript
const varSyntax1 = new RegExp(`[[${name}]]`, 'g');
const varSyntax2 = new RegExp(`$${name}`, 'g');

query = query.replace(varSyntax1, current.value);
query = query.replace(varSyntax2, current.value);
```

mrtraser: that's acts wrong actually, never mind

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+  let hasAggregates = false;
+  walk(parse(expression), obj => {
+    if (isSTSReference(obj) && obj.query.some(isSTSAggregateFilter)) {
+      hasAggregates = true;
+    }
+  });
+  return hasAggregates
+};
+
+export function flattenServerQueryFilters(items: STSFilter[]): StringMap {
```
mrtraser: mb we can have only function expression in a file (like function above)? Replace
```typescript
export function flattenServerQueryFilters(items: STSFilter[]): StringMap { ... }
```
with
```typescript
export const flattenServerQueryFilters = (items: STSFilter[]): StringMap => { ... }
```

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+
+const filterDeep = getFilterDeep(_);
+const compiledGrammar = Grammar.fromCompiled(grammar);
+
+export const formQueriesForExpression = async (
+  expr: string,
+  target: QueryTarget,
+  connector: Connector
+): Promise<DataQueryRequestItem[]> => {
+  if (isSimpleSyntheticExpression(expr)) {
```
mrtraser: can we get rid of this? cause here parser is run twice – inside `isSimpleSyntheticExpression ()` and at line 28. Mb we can just replace with `!getServerFilters(parsed).length;`

src/parser/ts/index.ts
polomani: you mean arrow functions for all exported members?

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+
+export type STSFilter = STSClientFilter | STSServerFilter;
+
+export type STSClientFilter = {
+  path: string;
+  filter: Exclude<FilterType, '='>;
+  value: string;
+};
+
+export type STSServerFilter = {
```
mrtraser: `STSRequestFilter` – mb it's a better name for such type?

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+    'on_error',
+  )
+}
+
+
+type TSResponseMap =  { [s: string]: TimeSeriesResponseItem };
+
+type StringMap = { [key: string]: string };
+
+type STSItem = (STSQueryItem[] | STSQueryItem)[] | STSQueryItem | STSFunction | WrappedConst;
```
mrtraser: we don't need `STSFunction` and `WrappedConst` here, right? They're included in `STSQueryItem`

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+) => {
+  if (isArray(route)) {
+    return route.map(r => walk(r, iterator));
+  }
+  const resArr = [];
+  const res = iterator(route);
+  if (res !== undefined) {
+    resArr.push(res);
+  }
+  if (isSTSFunction(route)) {
```
mrtraser: can we have a case where both `if` statements will be `true` (I can't see)? If no then we can replace this logic with `if .. else` and get rid of `resArr`

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+    let args: STSItem;
+    if(isMapFunction(route)) {
```
mrtraser: ternary operator?

src/parser/ts/index.ts
mrtraser: yep

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+};
+
+export const getServerFilters = (
+  route: STSQueryItem[] | STSFunction
+): StringMap[] => {
+  const responseArr: StringMap[] = [];
+  const filter = (_, __, parent) => isServerFilter(parent);
+  walk(route, obj => {
+    if (isSTSReference(obj) && !hasIdsFilter(obj)) {
+      const serverQuery = filterDeep(filterDeep(obj.query, filter), filter);
```
mrtraser: this can be replaced just with `filterDeep(obj.query, filter)`

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+  }
+  return res;
+}
+
+export const getClientFilters = (route: STSQueryItem[] | STSFunction): STSClientFilter[][] => {
+  const responseArr: STSClientFilter[][] = [];
+  const filter = (_, __, parent) => isClientFilter(parent) || isArray(parent.value);
+  walk(route, obj => {
+    if (isSTSReference(obj) && !hasIdsFilter(obj)) {
+      const clientQuery = filterDeep(filterDeep(obj.query, filter) || [], filter);
```
mrtraser: to avoid doing double deep filtering like:
```typescript
filterDeep(filterDeep(obj.query, filter) || [], filter);
```

we can replace : 
```typescript
const filter = (_, __, parent) => isClientFilter(parent) || isArray(parent.value);
```
with:
```typescript
const filter = (value, _, parent) => isClientFilter(parent) || (isArray(value) && value.some(isClientFilter));
```

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+      if (labelSrc) {
+        return getLabelWithInjectedProps(labelSrc, serie);
+      }
+      return `${serie.name || serie.externalId || serie.id}`;
+    }
+  });
+  return labelRes;
+};
+
+export const getLabelsForExpression = async (
```
mrtraser: In this function, we get timeseries again after `getDataQueryRequestItems` and always receive exactly the same data that in `getDataQueryRequestItems`. So all logic regarding labels can be simplified I believe. It's shouldn't be a part of this PR but probably we should keep this in mind

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+    const joinedStr = items.join(', ');
+    if (wrap) {
+      return isSTSFilterArr(val) ? `{${joinedStr}}`: `[${joinedStr}]`;
+    }
+    return joinedStr;
+  }
+  return JSON.stringify(val);
+}
+
+export const reverse = (
```
mrtraser: Nice one 👍 

But can we have a different name here? Smthig like `toSyntheticQuery` mb?

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+
+const getIdFilters = (obj: STSReference) => {
+  return obj.query.filter(isIdsFilter);
+};
+
+const hasIdsFilter = (obj: STSReference) => {
+  return getIdFilters(obj).length;
+};
+
+function isEqualsFilter(query: any): query is STSServerFilter  {
```
mrtraser: I like such kind of utils, it's much easier to read all conditions statements!

src/parser/ts/parser.unit.test.ts
```diff
@@ -0,0 +1,443 @@
...
+
+describe('permutations', () => {
+  const gen = generateAllPossiblePermutations;
+
+  it('one element', () => {
+    expect(gen([['a']])).toEqual([[['a']]]);
+  });
+
+  it('multiple elements', () => {
+    expect(gen([['a'], ['1', '2', '3'], ['å', 'æ']])).toEqual([
```
mrtraser: Nice) Now I at least understand what `generateAllPossiblePermutations` actually do 😅

polomani: yeah I know 😅
but at least it's abstracted a bit from other logic

src/parser/ts/index.ts
polomani: https://stackoverflow.com/questions/148857/what-is-the-opposite-of-parse
composeSyntheticQuery maybe?

mrtraser: `composeSyntheticQuery` sounds good

src/parser/ts/index.ts
polomani: no, not really. thing is synthetic query api endpoint doesn't support those filters.
if we rename it, we should rename STSClientFilter as well then, but I don't have ideas..

mrtraser: ok then, let's leave it as is

src/parser/ts/index.ts
polomani: oh, thanks. I can actually simplify types even more

src/parser/ts/index.ts
polomani: It doesn't seem like I ever use return functionality, I will just remove this chunk.

src/parser/ts/index.ts
polomani: hehe, you are fan of them) ok

src/parser/ts/index.ts
polomani: no, I have test for it, maybe change filter somehow?

src/parser/ts/index.ts
polomani: I wish it worked, but it breaks tests actually :(

src/parser/ts/index.ts
polomani: yep, agree, in fact I didn't do any smart logic for best performance.
we'll need to review on how many requests we send in this tab, we can lower down number by better logic and in some places - caching

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+}
+
+export const reverse = (
+  target: STSItem,
+  custom?: (item: STSItem) => string,
+  separateWith: string = ''
+): string => {
+  if (custom) {
+    const customRes = custom(target);
+    if (customRes !== undefined) {
```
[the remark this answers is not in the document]
polomani: could you elaborate? not following..

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+export const getIndicesOfMultiaryFunctionArgs = (route: STSQueryItem[] | STSFunction): number[] => {
+  const responseArr = [];
+  const argsIndices = [];
+  let index = 0;
+  walk(route, obj => {
+    if (isMultiaryFunction(obj) && obj.args.length === 1) {
+      argsIndices.push(...obj.args);
+    } else if (isSTSReference(obj) && !hasIdsFilter(obj)) {
+      if (argsIndices.includes(obj)) {
+        responseArr.push(index);
```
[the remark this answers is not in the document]
polomani: is it? but there is a few tests where response arr is not empty
```
it('works with composition of functions', () => {
    const res = indices(cloneDeep(funcComposition));
    expect(res).toEqual([2, 3]);
  });
```

src/query_ctrl.ts
polomani: we decided to do this as a separate task

src/parser/ts/index.ts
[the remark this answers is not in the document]
mrtraser: yes, I saw, but when I have been debugging it I received 2-dimensional array as argument (not like `funcComposition`)

src/parser/ts/index.ts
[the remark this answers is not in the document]
mrtraser: so the logic here if we have `custom` func we use it to reverse structure. But if `custom` returns `undefined` then logic inside it is wrong, right? Or if it returns undefined then we use default behavior as a fallback?

src/parser/ts/index.ts
mrtraser: oh, caching is smthing that should be also reviewed) But let's have sts perfectly working and deal with this after

src/parser/ts/index.ts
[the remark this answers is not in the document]
polomani: hmm, I see. I'll check

src/parser/ts/index.ts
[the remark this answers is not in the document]
polomani: > Or if it returns undefined then we use default behavior as a fallback?
exactly, custom is needed only for some specific cases, so we can use it function like map with condition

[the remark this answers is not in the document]
mrtraser: good then

src/parser/ts/index.ts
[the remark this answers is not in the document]
polomani: figured out that it works as expected

src/parser/ts/index.ts
polomani: fixed

src/parser/ts/index.ts
polomani: ffixed

src/parser/ts/index.ts
polomani: fixed

src/parser/ts/index.ts
polomani: fixed

src/parser/ts/index.ts
polomani: fixed

src/parser/ts/index.ts
polomani: decided to keep as it is

src/parser/ts/index.ts
polomani: decided to keep as it is

src/parser/ts/index.ts
```diff
@@ -0,0 +1,528 @@
...
+  return res;
+};
+
+export const convertExpressionToLabel = (
+  expression: string,
+  labelSrc: string,
+  tsMap: TSResponseMap
+) => {
+  const parsed = parse(expression);
+  const labelRes = reverse(parsed, item => {
```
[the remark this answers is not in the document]
polomani: fixed

PR #98, opened 2020-04-02
fix: array sts server filters

polomani:
basically new tests should be self explanatory
there was one bug, that didn't allow doing any array filtering.
also as a bonus added support for advanced grafana variables. (it allows variable formatting)

### DIFF
--- a/src/parser/ts/index.ts
+++ b/src/parser/ts/index.ts
@@ -166,14 +166,14 @@ export const hasAggregates = (
   return hasAggregates
 };
 
-export const flattenServerQueryFilters = (items: STSFilter[]): StringMap => {
-  return items.reduce((res, filter) => {
+export const flattenServerQueryFilters = (items: unknown[]): StringMap => {
+  return items.filter(isSTSFilter).reduce((res, filter) => {
     let value: any;
     if (isByIdsQuery(filter.value)) {
       value = filter.value.map(item => flattenServerQueryFilters([item]));
-    } else if(isSTSFilterArr(filter.value)) {
+    } else if (isSTSFilterArr(filter.value)) {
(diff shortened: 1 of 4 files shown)


src/parser/ts/index.ts
```diff
@@ -217,24 +217,30 @@ export const getIndicesOfMultiaryFunctionArgs = (route: STSQuery): number[] => {
...
     }
   });
-  return responseArr;
+  return extracted;
+};
+
+export const getServerFilters = (
+  route: STSQuery
+): StringMap[] => {
+  const filter = (_, __, parent) => isServerFilter(parent) || !isSTSFilter(parent);
```
mrtraser, line 236: Does `isServerFilter(parent)` not enough?

polomani: no. it skips `ts{prop=[...arr]}`, as parent just a regular array, not the object

src/parser/ts/parser.unit.test.ts
```diff
@@ -192,7 +202,12 @@ describe('nearley parser', () => {
 
   it('template variables', () => {
     const res = parse(`ts{id=[[asset]]}`);
-    expect(res).toEqual(STS([Filter('id', '$asset')]));
+    expect(res).toEqual(STS([Filter('id', '[[asset]]')]));
+  });
+
+  it('advanced template variable', () => {
+    const res = parse('ts{id=${asset:json}}');
+    expect(res).toEqual(STS([Filter('id', '${asset:json}')]));
```
mrtraser, line 210: nice!

src/parser/ts/index.ts
mrtraser: got it

PR #104, opened 2020-05-07
fix: parse escape char

mrtraser:


src/parser/common.ne
```diff
@@ -0,0 +1,46 @@
...
+
+strescape -> ["/bfnrt] {% id %}
+    | "u" [a-fA-F0-9] [a-fA-F0-9] [a-fA-F0-9] [a-fA-F0-9] {%
+    function(d) {
+        return d.join("");
+    }
+%}
+
+# space
+_ -> [\s]:*  {% null %}
```
polomani, line 46: `value` and `primitive` seem to be the same as well? can we move it here?

src/parser/events-assets/grammar.ne
```diff
@@ -1,11 +1,9 @@
 @preprocessor typescript
 
-@builtin "number.ne"
-@builtin "string.ne"
+@include "../common.ne"
```
polomani, line 3: `common.ne` doesn't need to be built separately? just import?

src/parser/ts/parser.unit.test.ts
```diff
@@ -250,6 +250,27 @@ describe('parse & reverse', () => {
...
+    String.raw`ts{name=~"\d some"}`,
+    String.raw`ts{name=~"\\d some"}`,
+    String.raw`ts{name=~"\\d \" some"}`,
+    String.raw`ts{name=~"\\d \\" some"}`,
+  ];
+  const outputs = [
+    STS([Filter('name', String.raw`\d some`, RegexEquals)]),
+    STS([Filter('name', String.raw`\\d some`, RegexEquals)]),
+    STS([Filter('name', String.raw`\\d " some`, RegexEquals)]),
+    STS([Filter('name', String.raw`\\d \" some`, RegexEquals)]),
```
polomani, line 264:
```suggestion
    STS([Filter('name', String.raw`\\d \" some`)]),
```
it's a default value

src/parser/common.ne
```diff
@@ -0,0 +1,46 @@
...
+    function(d) {
+        return JSON.parse("\""+d.join("")+"\"");
+    }
+%}
+
+sstrchar -> [\\] {% d => "\\" %}
+    | [^\\'\n] {% id %}
+    | "\\" strescape
+        {% function(d) { return JSON.parse("\""+d.join("")+"\""); } %}
+    | "\\'"
```
polomani, line 35: 31st and 35th line, do they produce ambiguity when we pass `\\'`?

src/parser/events-assets/parser.unit.test.ts
```diff
@@ -149,5 +149,64 @@ describe('Query parser', () => {
 
       expect(parsedWithSpaces).toEqual(parsedWithoutSpaces);
     });
+    it('should parse escape characters in single quoted string', () => {
+      const querySBS = String.raw`assets{metadata={key1!~'\value2 \d \. - \w+', key2!~"\value2 \d \. - \w+"}}`;
```
polomani, line 153: probably neat-picking, but maybe we could 
destruct string to get `raw` function? so then we don't need to be that verbose every time

```
const { raw: r } = String;
```
then it will look like in python)))

src/parser/common.ne
mrtraser: yep, should be also in a `common.ne`

src/parser/events-assets/grammar.ne
mrtraser: yes, `common.ne` will be included in the dictionary while building

src/parser/common.ne
mrtraser: No, they don't, just checked

src/parser/events-assets/parser.unit.test.ts
mrtraser: haha, let's try)

src/parser/common.ne
polomani: awesome

src/parser/ts/grammar.ne
```diff
@@ -139,32 +126,18 @@ type -> "ts" {% id %}
...
-equals -> "=" {% id %}
-not_equals -> "!=" {% id %}
-prop_name -> [A-Za-z0-9_]:+ {% join %}
-
 string -> sqstring {% id %}
   | dqstring {% id %}
   | variable {% id %}
-number -> decimal {% id %}
+regexp_string -> sqregexp {% id %}
+  | dqregexp {% id %}
```
polomani, line 133:
```suggestion
  | dqregexp {% id %}
  | variable {% id %}
```
right? so it behaves like before

polomani, line 133:
```suggestion
  | dqregexp {% id %}
  | variable {% id %}
```
right? so it behaves like before

mrtraser: I'll add another test for it

mrtraser:
so yes, no ambiguity anymore 😃

src/parser/ts/grammar.ne
mrtraser: done

PR #105, opened 2020-05-14
fix: default values of aggregates in STS

mrtraser:
Add default values of aggregation and granularity for synthetic timeseries if they are not provided.

### DIFF
--- a/src/cdfDatasource.ts
+++ b/src/cdfDatasource.ts
@@ -24,11 +24,11 @@ import {
 import { get, cloneDeep } from 'lodash';
 import { ms2String } from './utils';
 import { Connector } from './connector';
-import { getLabelsForExpression, hasAggregates } from './parser/ts';
+import { getLabelsForExpression } from './parser/ts';
 import { getRange } from './datasource';
 import { TimeSeries } from '@grafana/ui';
 import { appEvents } from 'grafana/app/core/core';
-import { failedResponseEvent, CacheTime, DATAPOINTS_LIMIT_WARNING } from './constants';
+import { datapointsLimit, failedResponseEvent, CacheTime, DATAPOINTS_LIMIT_WARNING } from './constants';
 
 const { Asset, Custom, Timeseries } = Tab;
 
@@ -39,8 +39,7 @@ export function formQueryForItems(
 ): DataQueryRequest {
   const [start, end] = getRange(options.range);
   if (tab === Custom) {
-    const isAggregated = items.some(({ expression }) => hasAggregates(expression));
-    const limit = calculateDPLimitPerQuery(isAggregated, items.length);
+    const limit = calculateDPLimitPerQuery(items.length);
     return {
       items: items.map(({ expression }) => ({ expression, start, end, limit })),
     };
@@ -53,7 +52,7 @@ export function formQueryForItems(
       granularity: granularity || ms2String(options.intervalMs),
     };
   }
-  const limit = calculateDPLimitPerQuery(isAggregated, items.length);
+  const limit = calculateDPLimitPerQuery(items.length);
   return {
     ...aggregations,
     end,
@@ -63,8 +62,8 @@ export function formQueryForItems(
   };
 }
 
-function calculateDPLimitPerQuery(hasAggregates: boolean, queriesNumber: number) {
-  return Math.floor((hasAggregates ? 10_000 : 100_000) / Math.min(queriesNumber, 100));
+function calculateDPLimitPerQuery(queriesNumber: number) {
+  return Math.floor(datapointsLimit / Math.min(queriesNumber, 100));
 }
 
 export function formQueriesForTargets(
--- a/src/cdfDatasource.ts
+++ b/src/cdfDatasource.ts
@@ -28,7 +28,7 @@ import { getLabelsForExpression } from './parser/ts';
 import { getRange } from './datasource';
 import { TimeSeries } from '@grafana/ui';
 import { appEvents } from 'grafana/app/core/core';
-import { datapointsLimit, failedResponseEvent, CacheTime, DATAPOINTS_LIMIT_WARNING } from './constants';
+import { failedResponseEvent, CacheTime, DATAPOINTS_LIMIT_WARNING } from './constants';
 
 const { Asset, Custom, Timeseries } = Tab;
 
@@ -52,7 +52,7 @@ export function formQueryForItems(
       granularity: granularity || ms2String(options.intervalMs),
     };
   }
-  const limit = calculateDPLimitPerQuery(items.length);
+  const limit = calculateDPLimitPerQuery(items.length, isAggregated);
   return {
     ...aggregations,
     end,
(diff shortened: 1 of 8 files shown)


src/parser/ts/index.ts
```diff
@@ -15,9 +21,19 @@ const compiledGrammar = Grammar.fromCompiled(grammar);
 export const formQueriesForExpression = async (
   expression: string,
   target: QueryTarget,
-  connector: Connector
+  connector: Connector,
+  options: QueryOptions
```
polomani, line 25: maybe pass just interval?

src/parser/ts/index.ts
```diff
@@ -48,6 +64,32 @@ const NoTimeseriesFound = (filter: StringMap, expr: string) => {
   );
 };
 
+/**
+ * Injects default values of aggregate and granularity (if it is not provided).
+ * Mutates <query> properties array with default aggregate or granularity values.
```
polomani, line 69: it doesn't mutate, it returns a copy

polomani: well, internally it mutates, but it's not visible outside anyway

src/parser/ts/index.ts
```diff
@@ -48,6 +64,32 @@ const NoTimeseriesFound = (filter: StringMap, expr: string) => {
...
+  parsedData: STSQuery,
+  defaultValues: { aggregate: string; granularity: string }
+) => {
+  const result = cloneDeep(parsedData);
+
+  walk(result, obj => {
+    if (isSTSReference(obj)) {
+      Object.keys(defaultValues).forEach(key => {
+        if (findIndex(obj.query, ['path', key]) === -1) {
+          obj.query.push(STSFilter(key, defaultValues[key]));
```
polomani, line 84: ;)

src/parser/ts/parser.unit.test.ts
```diff
@@ -449,3 +450,54 @@ describe('check if expression has aggregates', () => {
...
+    const withAggregatesAndGranularity = injectAggregatesToParsed(
+      stsWithAggregateAndGranularity,
+      defaults
+    );
+
+    expect(withAggregatesAndGranularity).toEqual(stsWithAggregateAndGranularity);
+  });
+  it('should not add default values in case of default aggregate is none', () => {
+    const sts = STS([Filter('name', 'name')]);
+    const stsWithInjected = injectAggregatesToParsed(sts, { aggregate: 'none', granularity: '2h' });
```
polomani, line 499: this should be removed now

src/partials/query.editor.html
```diff
@@ -142,11 +142,24 @@
         </span>
       </info-popover>
     </div>
+    <div class="gf-form">
+      <label class="gf-form-label query-keyword fix-query-keyword">Aggregation</label>
+      <select class="gf-form-input" ng-model="ctrl.target.aggregation" ng-init="ctrl.getInitAggregate()" ng-change="ctrl.refreshData()">
+        <option ng-repeat="fn in ctrl.customTabAggregation" value="{{fn.value}}">{{fn.name}}</option>
+      </select>
+    </div>
+    <div class="gf-form" ng-if="ctrl.target.aggregation && ctrl.target.aggregation !== 'none'">
```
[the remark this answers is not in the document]
mrtraser: yes, and here is used to show/hide when aggregation is selected or not

src/parser/ts/index.ts
mrtraser: yep, will change

src/parser/ts/index.ts
mrtraser: I've mentioned it cause implementation is a bit step away of a functional approach

src/parser/ts/parser.unit.test.ts
mrtraser: why? to set aggregation to `none` is the only way to fetch real data instead of aggregated

polomani: I though we could operate with undefined/null value instead.
but ok, this approach works as well

src/cdfDatasource.ts
```diff
@@ -52,7 +51,7 @@ export function formQueryForItems(
       granularity: granularity || ms2String(options.intervalMs),
     };
   }
-  const limit = calculateDPLimitPerQuery(isAggregated, items.length);
+  const limit = calculateDPLimitPerQuery(items.length);
```
polomani, line 54: for non-aggregated values in normal timeseries limit is still 100k
so no need to change here

src/parser/ts/index.ts
```diff
@@ -15,9 +21,19 @@ const compiledGrammar = Grammar.fromCompiled(grammar);
...
   expression: string,
   target: QueryTarget,
-  connector: Connector
+  connector: Connector,
+  defaultInterval: string
 ): Promise<DataQueryRequestItem[]> => {
-  const parsed = parse(expression);
+  const rawParsed = parse(expression);
+  const defaultGranularity = target.granularity || defaultInterval;
+  const defaultAggregation = target.aggregation;
```
polomani, line 29: maybe we just pass all this stuff (`target. granularity, granularity.aggregatetion, interval`) to the `injectAggregatesToParsed ` and hide logic there?
same with `defaultAggregation === 'none'`

polomani: end call it `enrichWithDefaultAggregates` or something

polomani: so we don't bloat the main function

mrtraser: hmm, it makes sense, I'll redo this as well

src/parser/ts/parser.unit.test.ts
mrtraser: I did this in this way cause we have `none` value in 1st tab, just to keep consistency when we're switching tabs

src/cdfDatasource.ts
```diff
@@ -63,8 +62,8 @@ export function formQueryForItems(
   };
 }
 
-function calculateDPLimitPerQuery(hasAggregates: boolean, queriesNumber: number) {
```
polomani, line 66: for regular ts limit is still different for aggregated/not aggregated dps

src/parser/ts/index.ts
polomani: I believe it just gives wrong assumptions to the user, as what matters is inputs/outputs, not internal logic.

src/parser/ts/parser.unit.test.ts
```diff
@@ -492,3 +493,76 @@ describe('parse unary "-" operator', () => {
     })
   );
 });
+describe('check if default aggregation and granularity will be added', () => {
+  const defaults = { aggregate: 'average', granularity: '1h' };
+  const defaultOnlyAggregate = { aggregate: 'average', granularity: '' };
+  const defaultInterval = '6h';
+  it('should add default aggregate and granularity', () => {
+    const stsWithoutAggregation = STS([Filter('name', 'name')]);
+    const withoutAggregates = enrichWithDefaultAggregates(
```
polomani, line 502:
```suggestion
    const withAggregates = enrichWithDefaultAggregates(
```
?

mrtraser: fixed

src/cdfDatasource.ts
mrtraser: ah, right, nice catch!

PR #106, opened 2020-05-14
fix: caching requests

polomani:


jest.config.js
```diff
@@ -8,5 +8,6 @@ module.exports = {
   moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
   moduleNameMapper: {
     "grafana/app/core/core": "<rootDir>/src/__mocks__/grafana/core/core.ts"
-  }
+  },
+  collectCoverageFrom: ["src/**/{!(grammar),}.ts"]
```
mrtraser, line 12: nice!

src/cdfDatasource.ts
```diff
@@ -156,12 +160,14 @@ export async function getTimeseries(
...
         method,
         path: `/timeseries/byids`,
+        cacheTime: '1h',
       });
     } else {
       items = await connector.fetchAndPaginate({
         data,
         method,
         path: `/timeseries/list`,
+        cacheTime: '1h',
```
mrtraser, line 170: can we have lower value here, cause during `1h` some ts might be added that will match provided filters

polomani: fixed, now it's 151 seconds (prime number so they don't overlap)

PR #108, opened 2020-05-20
feat: add unary minus parser support

mrtraser:


src/parser/ts/grammar.ne
```diff
@@ -108,12 +123,12 @@ function -> unary br arithmethicElements BR {% extractFunction %}
...
 commaSeparatedElements -> compositeElement (comma compositeElement):* {% extractCommaSeparatedArray %}
-arithmethicElements -> element (operator element):+ {% extractOperationsArray %}
+arithmeticElements -> (unary_operator):? element (operator (unary_operator):? element):+ {% extractOperationsArray %}
 
 map_func_args -> compositeElement "," array "," array comma number {% extractMapFuncArgs %}
 
-compositeElement -> element {% id %}
-  | arithmethicElements {% id %}
+compositeElement -> (unary_operator):? element {% extractUnaryOperator %}
+  | arithmeticElements {% id %}
```
polomani, line 131: it seems like a bit too much of grammar / logic for this minor feature.
can we try to addd it just in one place?

like in `function` for instance, since it's used as an "element" everywhere?

mrtraser: it will produce additional ambiguity if I'll add this logic to `function` instead

src/parser/ts/grammar.ne
```diff
@@ -92,14 +95,26 @@ const extractFunction = ([func, br, args, BR]) => {
...
 %}
 
 query -> _ trimmed _ {% d => d[1] %}
 
 trimmed -> compositeElement {% id %}
   | function {% id %}
 
-function -> unary br arithmethicElements BR {% extractFunction %}
+function -> br unary_operator element BR {% brakedUnaryOperator %}
+  | unary br arithmeticElements BR {% extractFunction %}
```
polomani, line 117: `unary br arithmeticElements BR` already includes element in brackets `(element)`
doesn't it work without line 116?

mrtraser: no, cause it fails to parse `(-ts{})`

PR #112, opened 2020-05-27
fix: per datapoint error handling

polomani:
Screenshot 2020-05-27 at 14 53 31

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -42,7 +42,7 @@ import {
 import { Connector } from './connector';
 import { TimeRange } from '@grafana/ui';
 import { ParsedFilter, QueryCondition } from './parser/types';
-import { datapointsLimitWarningEvent, failedResponseEvent } from './constants';
+import { datapointsWarningEvent, failedResponseEvent } from './constants';
 
 const { Asset, Custom, Timeseries } = Tab;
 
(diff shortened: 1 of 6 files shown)


src/datasource.ts
```diff
@@ -356,28 +356,37 @@ export function getRange(range: TimeRange): Tuple<number> {
...
+      (datapoints as [])
+        .map(({ error }) => error)
+        .filter(Boolean)
+        .forEach(error => {
+          calculationErrors.add(error);
+        });
+    });
+
+    calculationErrors.forEach(message => {
+      appEvents.emit(datapointsWarningEvent, { refId, warning: message });
```
mrtraser, line 389: mb we can send only one event here but with the array of warnings?

src/datasource.ts
```diff
@@ -356,28 +356,37 @@ export function getRange(range: TimeRange): Tuple<number> {
...
+    const hasMorePoints = items.some(({ datapoints }) => datapoints.length >= limit);
+    const warning =
+      'Datapoints limit was reached, so not all datapoints may be shown. Try increasing the granularity, or choose a smaller time range.';
+
+    if (hasMorePoints) {
+      appEvents.emit(datapointsWarningEvent, { refId, warning });
     }
-  );
+
+    const calculationErrors = new Set();
```
mrtraser, line 378: can we split logic into separate functions (`checkLimits`, `checkCalculationErrors` ) ?

src/datasource.ts
polomani: fixed, but I've sent it as one string
should be ok as well

src/datasource.ts
polomani: fixed

PR #125, opened 2020-08-19
fix(custom-query): properly display pure text labels [CDF-8145]

polomani:
Custom query: display legend labels as a plain text instead of expression.
Example for query `ts{id=1} + ts{id=2} + 1`:
- user label `{{name}}` will result to `timeseries_name_1 + timeseries_name_2 + 1` label (as it used to work)
- user label `Cool custom name` will result to `Cool custom name` label (since no specific timeseries props referenced)


tengelskar (approved):
Looks good!

PR #126, opened 2020-08-24
fix: misleading error message in custom query [CDF-6622] [CDF-7537]

polomani:
Now it shows the actual error message instead.

- also some refactor
- introduced a bug here https://github.com/cognitedata/cognite-grafana-datasource/pull/125 Added tests and fixed it now
- upraded prettier since it didn't work with?.optional.?chaining

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -35,11 +35,12 @@ import {
   formQueriesForTargets,
   getTimeseries,
   reduceTimeseries,
-  formMetadatasForTargets,
   promiser,
   getLimitsWarnings,
   getCalculationWarnings,
   datapointsPath,
+  stringifyError,
+  getLabelsForTarget,
 } from './cdfDatasource';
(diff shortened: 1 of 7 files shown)


dbrattli (approved):
LGTM!

src/datasource.ts
```diff
@@ -104,21 +105,27 @@ export default class CogniteDatasource {
...
+    const queries = formQueriesForTargets(queryData, options);
+    const metadata = await Promise.all(
+      queryData.map(async ({ target, items }) => {
+        let labels = [];
+        try {
+          labels = (await getLabelsForTarget(target, items, this.connector)).map(label =>
+            this.replaceVariable(label, options.scopedVars)
+          );
+        } catch (err) {
+          handleError(err, target.refId);
```
mrtraser, line 124: nice here, much better than have it inside `cdfDatasource`

PR #127, opened 2020-08-24
fix: default granularity is not less than 1s [CDF-8306]

polomani:


tengelskar (approved):
Looks good, consider documenting the behavior of ms2String since it returns max(1000, milliseconds) which isn't obvious from the function name.

PR #131, opened 2020-09-22
ci: remove jenkins, add gh-action

polomani:
I also added a jira ticket regarding docker image generation part. We can do that later


dbrattli (approved):
LGTM

PR #132, opened 2020-10-05
React port

dbrattli:


src/types.ts
```diff
@@ -29,8 +29,26 @@ export enum Tab {
...
+  target: '',
+  // type: 'timeserie',
+  aggregation: 'average',
+  granularity: '',
+  label: '',
   tab: Tab.Timeseries,
+  expr: '',
+  assetQuery: {
+    target: '',
+    old: undefined,
```
polomani, line 49: old is not needed anymore

package.json
```diff
@@ -39,7 +39,7 @@
     "jest": "^26",
     "lint-staged": "^8.1.1",
     "ms": "^2.1.2",
-    "nearley": "^2.19.1",
+    "nearley": "^2.19",
```
polomani, line 55: this doesn't do anything, or is it some logic behind it?

webpack.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     fs: "empty"
   },
   context: __dirname + "/src",
-  entry: "./module.ts",
+  entry: "./module.tsx",
```
polomani: it shouldn't be a `tsx` file as we don't have any markup in it `ts` is enough

src/__tests__/cdfDatasource.test.ts
```diff
@@ -74,13 +69,15 @@ describe('CDF datasource', () => {
 
   describe('promiser', () => {
     it('should return failures and successes', async () => {
-      const queries = [0, 1, 2, 3];
-      const metadatas = ['a', 'b', 'c', 'd'];
+      const queries = [0, 1, 2, 3] as unknown as CDFDataQueryRequest[];
+      const metadatas = ['a', 'b', 'c', 'd'] as unknown as ResponseMetadata[];
```
polomani, line 73: this is very bad

dbrattli: Yes it's a bad test. We will not refactor the tests as part of the port to React. At least this shows how bad it already was.

src/types.ts
dbrattli: Fixed

src/__tests__/cdfDatasource.test.ts
polomani: the function used to be generic.
so it worked with anything before,
but then you guys decided to lock it to this implementation :)

dbrattli: Aha, yes that's right. I still think it's a good choice. The `promiser` function should not need to care about Query, Metadata or Responses so we should refactor this. It's too much detail for a function that simply takes a list of promise generating functions and returns a single promise with accumulated result.

polomani: exactly!

dbrattli: I will try to refactor this

src/types.ts
```diff
@@ -157,9 +158,35 @@ export type FailResponse = {
...
+
+  constructor(v: T) {
+    this.value = v;
+  }
+
+  isOk: boolean = true;
+  isErr: boolean = false;
+}
+
+export class Err<T, E> implements Result<T, E> {
```
polomani, line 176: oh no, classes in js

src/cdf/client.ts
```diff
@@ -214,30 +218,17 @@ function datapoint2Tuple(
...
-  };
+export async function concurrent<TQuery, TResult, TError>(
+  queries: TQuery[],
+  queryProxy: (query: TQuery) => Promise<Result<TResult, TError>>
+): Promise<Responses<TResult, TError>> {
+  const later = _.map(queries, (query) => queryProxy(query));
+  const results = await Promise.all(later);
+  const [oks, errs] = _.partition(results, (res) => res.isOk);
+  const succeded = _.map(oks, (x: Ok<TResult, TError>) => x.value);
+  const failed = _.map(errs, (x: Err<TResult, TError>) => x.error);
```
polomani, line 229: why not just `errs.map(({ error }) => error)`?

dbrattli: Fixed, thanks!

src/plugin.json
```diff
@@ -55,7 +55,7 @@
       "headers": [
         {
           "name": "api-key",
-          "content": "{{.SecureJsonData.cogniteDataPlatformApiKey}}"
+          "content": "{{.SecureJsonData.apiKey}}"
```
polomani, line 58: need to check if it's a breaking change

polomani: it was, fixed that

polomani:
We need to check if this PR works on grafana 6. If not - bump required grafana version

src/datasource.ts
```diff
@@ -1,95 +1,112 @@
...
     if (queryTargets.length) {
       try {
         const { failed, succeded } = await this.fetchTimeseriesForTargets(queryTargets, options);
         handleFailedTargets(failed);
         showWarnings(succeded);
         responseData = reduceTimeseries(succeded, getRange(options.range));
       } catch (error) {
-        console.error(error); // not sure it ever happens
+        /* eslint-disable-next-line no-console  */
+        console.error(error); // TODO: use app-events or something
```
dbrattli, line 109: I was worred about this one, but it's fine to revert as the fix does not need to be part of this PR.

dbrattli:
Looks good. Great work @polomani and contributors!

polomani (approved):
LGTM thanks for the work, everyone!

tengelskar (commented):
This is so cool! Great work!

PR #133, opened 2020-10-21
Add templates query

mortenhillbom:


polomani:
Hi @mortenhillbom 
fyi we are working on a huge change to the codebase, moving everything to react here https://github.com/cognitedata/cognite-grafana-datasource/pull/132
that would basically make all angular stuff incopatible.
what are you intentions in this draft here?

mortenhillbom:
> Hi @mortenhillbom
> fyi we are working on a huge change to the codebase, moving everything to react here #132
> that would basically make all angular stuff incopatible.
> what are you intentions in this draft here?

Ooh, thanks for letting me know!
Is #132 "stable" enough now that I can start porting this code as well and rebasing on #132, or should I wait until it is closer to being finished and merged?

The intention of this PR is to validate the complete user journey for the new templates service we are building :) where grafana integration is a highly requested feature. The templates service itself is still a work in progress, so initially our wish is a way to share this with just a few people who want to try it out.
However in the longer term we would like to have templates functionality be a part of the main plugin :)

polomani:
I don't have a strong opinion if this should be based on #132, @dbrattli what do you think?

src/templatesDatasource.ts
```diff
@@ -0,0 +1,280 @@
...
+        dataPathArray.push(trimmed);
+      }
+    }
+    if (!dataPathArray) {
+      throw 'data path is empty!';
+    }
+    return dataPathArray;
+  }
+
+  async query(options: DataQueryRequest<TemplateQuery>): Promise<DataQueryResponse> {
```
dbrattli, line 143: Too much nesting here: for -> for -> for -> if -> for -> if

Tapped: Yeah we were unsure if we want to refactor it now, or later. Its copied pasted from a GraphQL Grafana plugin: https://github.com/fifemon/graphql-datasource/blob/master/src/DataSource.ts#L130

Tapped: Also @mortenhillbom since we are copy pasting code, we need to make sure to handle the license correctly. However, we might as well just write this ourself, which will most likely be considered our own work.

PR #137, opened 2020-11-13
feat: timeseries external ids

polomani:
I cannot say I am happy with the code, but I didn't find any better abstraction to make ui work the same way.

There are few different cases, and this is covering them at least, tested manually.


dbrattli (approved):
Looks good. Great work!

tengelskar (approved):

PR #144, opened 2020-11-23
fix: variable input in assets tab

polomani:
We lost some capabilities with the recent update.
Not possible to see/input $variables in asset tab. 

Before the fix:
Screenshot 2020-11-23 at 10 06 46

After:
Screenshot 2020-11-23 at 10 06 30


dbrattli (approved):
Looks good!

PR #145, opened 2020-12-04
feat: time series latest value

polomani:
Enables "latest data point" feature in "timeseries" tab

Screenshot 2021-01-11 at 10 12 46

*REFACTOR:*
Due to the code being too inconveniet for unit testing, I rewrote some logic around variable replacing.
Now all the `variable.replace(str, ...)` login is moved in one place, it happens before any other business logic.

### DIFF
--- a/src/__tests__/cdf.client.test.ts
+++ b/src/__tests__/cdf.client.test.ts
@@ -7,7 +7,7 @@ import {
 } from '../cdf/client';
 import { getDataqueryResponse, getMeta } from './utils';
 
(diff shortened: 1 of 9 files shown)


polomani:
- fix the "label" input
- make "$__to" a default

polomani:
- fix the `label` input dissapearance
- made time range end as a default `before` prop
- removed `before` input
- fixed css

tengelskar (approved):
Looks good!

PR #172, opened 2022-02-10
Forward errors to Grafana UI

Tapped:
This makes errors visible in the corner as the image below shows, which how I expect errors in a datasource to be shown.

image


polomani:
included in https://github.com/cognitedata/cognite-grafana-datasource/pull/173

PR #176, opened 2022-03-04
Relationship support

KissSpinobleZsolt:
Support for the Relationships


polomani (changes requested):
Thanks for the PR.
First of all, I would like to ask you to revert all formatting changes, as it's hard to see where it's just format vs actual code changes. You can submit formatting as a separate PR as well.

Also, `yarn.lock` has changed but not `package.json`?

I've put some comments about the code style as well.

polomani:
Screenshot 2022-03-07 at 12 31 38

I am not able to visualise nodes

src/components/RelationshipsListTab.tsx
```diff
@@ -0,0 +1,74 @@
+import { MultiSelect } from '@grafana/ui';
```
polomani, line 1: we use camelcase for other file names

src/__tests__/RelationshipsListTab.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
polomani, line 1: we use camelcase for other file names in the project

polomani (commented):
Looks better then before,

I left some comments, that needed to be fixed to make the code a bit less messy.

Also there is a bug when I don't see selected labels and dataset after page reload.
Screenshot 2022-03-16 at 13 06 53

yarn.lock
```diff
@@ -2,83 +2,19 @@
 # yarn lockfile v1
 
 
-"@antv/adjust@~0.1.0":
```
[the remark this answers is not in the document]
KissSpinobleZsolt: It may come from a node-version mismatch, maybe I started or I ran yarn with other node version then v14 what I'm using for development

src/__tests__/datasource.test.ts
```diff
@@ -249,7 +253,9 @@ describe('Datasource Query', () => {
       { id: 456, externalId: 'Timeseries456', description: 'test timeseriesB' },
       { id: 789, externalId: 'Timeseries789', description: 'test timeseriesC' },
     ]);
-
+    /* TODO
+    intergrate the Realationship query
+    */
```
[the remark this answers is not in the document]
KissSpinobleZsolt: sorry for getting todo.... I hade some problem with git merge. Usually I don't let TODO's in.

src/utils.ts
```diff
@@ -40,3 +42,156 @@ export const checkFilter = <T>(obj: T, { path, filter, value }: ParsedFilter): b
...
+  return { sourceMeta, targetMeta };
+};
+export const generateNodesAndEdges = (
+  realtionshipsList: CogniteRelationshipResponse[],
+  refId: any
+) => {
+  const iterrer = getMetaKeys(realtionshipsList);
+  const nodes = nodesFrame(iterrer, refId);
+  const edges = edgesFrame(refId);
+  map(
```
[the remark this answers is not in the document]
KissSpinobleZsolt: I use instead lodash map, JS map, I don't prefer for loop, but if you'd like I will use it.

src/datasource.ts
```diff
@@ -454,14 +538,17 @@ export default class CogniteDatasource extends DataSourceApi<
...
+      const { tab, assetQuery, eventQuery, relationsShipsQuery } = target;
       switch (tab) {
         case Tab.Event:
           return eventQuery?.expr;
         case Tab.Asset:
           return assetQuery?.target;
         case Tab.Custom:
           return target.expr;
+        case Tab.Relationships: {
+          return relationsShipsQuery;
```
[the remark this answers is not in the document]
KissSpinobleZsolt: It has to be always true, we need the unfiltered list also.

src/components/resourceSelect.tsx
```diff
@@ -50,7 +45,7 @@ export function ResourceSelect(props: {
 
   const fetchDropdownResource = async (id: IdEither) => {
     try {
-      const [res] = await fetchSingleResource(id);
+      const [res] = !isEmpty(Object.values(id).join('')) && (await fetchSingleResource(id));
```
polomani, line 48:
```
IdEither = {
      id: number;
    }
  | {
      externalId: string;
    };
```
1. `IdEither` shouldn't be empty, if it is then it means types aren't correct
2. This is a bit hackish check anyway, could be at least `Object.values(id).filter(Boolean)`

src/datasource.ts
```diff
@@ -261,6 +260,14 @@ export default class CogniteDatasource extends DataSourceApi<
     const timeFrame = {
       activeAtTime: { min: start, max: end },
     };
+    const filtered = each(targets, ({ tab }) => tab === Tab.Relationships);
+    if (filtered.length) {
+      return Promise.all(
+        targets.map(async (target) => {
+          return this.createRelationshipsNode(target.relationsShipsQuery, target.refId);
```
polomani, line 267: what is this code doing here? this function is made for events. no relationships are expected here

src/datasource.ts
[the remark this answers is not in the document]
polomani: then I'd strongly suggest returning just `true`

src/datasource.ts
```diff
@@ -588,6 +802,9 @@ export async function getDataQueryRequestItems(
 }
 
 function groupTargets(targets: CogniteQuery[]) {
-  const [eventTargets, tsTargets] = partition(targets, ({ tab }) => tab === Tab.Event);
+  const [eventTargets, tsTargets] = partition(
+    targets,
+    ({ tab }) => tab === Tab.Event || tab === Tab.Relationships
```
polomani, line 807: this function needs to be rewritten to support 3 groups, something like:

```
const { events, timeseries, relationships } = groupBy(targets,  ({ tab }) => (tab || Tab.Timesseries).toLowerCase() ))
return { eventTargets: events, tsTargets: timeseries, relationshipsTargets: relationships };
```

polomani: fyi `Tab.Timesseries` is the default, so undefined should appear as timeseries as well

src/cdf/types.ts
```diff
@@ -139,15 +139,68 @@ export type Items<T = any> = {
...
+  name?: string;
+  description?: string;
+  metadata?: any;
+  writeProtected?: boolean;
+  id?: number;
+  createdTime?: number;
+  lastUpdatedTime?: number;
+}
+
+interface CogniteRealtionshipAsset {
```
polomani, line 167:
```suggestion
interface CogniteRelationshipAsset {
```

polomani (commented):
I've put some more comments.
You need to remove the hack where relationships are being injected into events logic.

src/components/resourceSelect.tsx
KissSpinobleZsolt: without this ` !isEmpty(Object.values(id).join('')) `, every time as you navigate on timeseries tab will be triggered a call with the `id: ''` , I think is coming from the `defaultQuery` is this wanted?

src/datasource.ts
KissSpinobleZsolt: >1. ` case Tab.Relationships: { return relationsShipsQuery; }`
here I can't return a bool value. I have to return the object holding the selected labels and datasets by default are empty arrays see `defaultRelationshipsQuery: RelationshipsQuery`

>2. for `const [eventTargets, tsTargets] = partition(targets, ({ tab }) => tab === Tab.Event);` see next comment with: `fetchEventTargets && groupTargets`

src/datasource.ts
KissSpinobleZsolt: Maybe I didn't understand well the "ask" before to move the code and integrate into somewhere else. I'm looking forward for a better integration.
the 2 "hot" points : `fetchEventTargets && groupTargets` will be reanalyzed and reimplemented in a way to don't conflict with other tabs, and to keep:  the `Tab.Timesseries` as the default, so undefined should appear as timeseries as well.

src/components/resourceSelect.tsx
polomani: got you!

src/datasource.ts
polomani: @KissSpinobleZsolt yes, essentially you need your own `fetchRelationshipTargets` that will take `relationshipTargets` result from `groupTargets` call

polomani: then in the main handler you will only have
```
 const relationshipsResults = await this.fetchRelationshipsTargets(relationshipTargets);
 // ....
 responseData = [...reduceTimeseries(succeded, timeRange), ...eventResults, ...relationshipsResults];
```

src/datasource.ts
polomani: >  here I can't return a bool value. I have to return the object holding the selected labels and datasets by default are empty arrays see defaultRelationshipsQuery: RelationshipsQuery

could you elaborate? I don't see the reasoning

src/datasource.ts
KissSpinobleZsolt: @polomani yes, I'm hard thinking on it to integrate, it bothers me 2 things. 
1. I have to map array of query's, but I may give back only one array of [nodes, edges] .. here my issue is the reducer function n the end of the calls (I'm hardly trying to fix this issue)
2. at the relationshipslistTab  I'm getting 
      `[ERROR] Parser: Unexpected end of input:
      
      ^` 
    and I don't know from where it is, it happens only if timeseries are also selected.

src/datasource.ts
KissSpinobleZsolt: I fixed it, I've put a constant true and now it works!!!

src/datasource.ts
polomani: > [ERROR] Parser: Unexpected end of input:

make sure that targets are grouped correctly, maybe `relationshipslist` target somehow appears in assets/events handler

KissSpinobleZsolt: Its not about grouping, grouping is good, the problem is the always true relationshipsQuery, if I select a value on a different tab then Relationship the relationship is also called messing the data with unwanted [nodes, edges]. now I fixed it in the call no to be called if the tab is not the right on, in our case the Tab.Relationship

polomani:
@KissSpinobleZsolt this can be closed, right?

polomani:
closed in favor of the newer PR

PR #180, opened 2022-05-06
assetTab  relationships support

KissSpinobleZsolt:


polomani (commented):
What's the purpose of "includeSubTimeseries"?
I seem to break dashboards created by the older version, by having the flag value to be undefined by default?

Screenshot 2022-05-10 at 09 28 07

src/datasource.ts
```diff
@@ -577,11 +590,35 @@ async function findAssetTimeseries(
...
   // since /dataquery can only have 100 items and checkboxes become difficult to use past 100 items,
   //  we only get the first 100 timeseries, and show a warning if there are too many timeseries
   const limit = 101;
-  const ts = await getTimeseries({ filter, limit }, connector);
+  let ts = [];
+  if (assetQuery.includeSubTiemseries) {
+    const tS = await getTimeseries({ filter, limit }, connector);
+    if (!isEmpty(tS))
+      tS.map(({ id, isStep, createdTime, lastUpdatedTime }) =>
+        ts.push({ id, isStep, createdTime, lastUpdatedTime, selected: true })
```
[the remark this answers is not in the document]
KissSpinobleZsolt: I just reused the original function. if you say it is not used anymore I can take it out

src/datasource.ts
```diff
@@ -577,11 +590,35 @@ async function findAssetTimeseries(
...
+        ...filterdataSetIds(dataSetIds),
+        ...filterExternalId(sourceExternalIds),
+        ...timeFrame,
+      },
+      limit,
+      connector
+    );
+    if (!isEmpty(relationshipsList))
+      relationshipsList.map(({ target: { id, isStep, createdTime, lastUpdatedTime } }) =>
+        ts.push({ id, isStep, createdTime, lastUpdatedTime, selected: true })
```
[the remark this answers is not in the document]
KissSpinobleZsolt: I just reused the original function. if you say it is not used anymore I can take it out

muradsater:
> What's the purpose of "includeSubTimeseries"?

Should be set to true by default to not break existing dashboard. When you have relationships you might not want to show sub-timeseries in some dashboards, e.g. if you are using activeAtTime for you relationsships.

polomani:
> Should be set to true by default to not break existing dashboard

@muradsater yes, that was my biggest concern here. But I think we managed to fix the compatibility now with @KissSpinobleZsolt.

On that note, I also suggest adding some "hints" to explain what precisely each toggle does.

src/datasource.ts
```diff
@@ -577,11 +590,33 @@ async function findAssetTimeseries(
...
+        ...timeFrame,
+      },
+      limit,
+      connector
+    );
+    if (!isEmpty(relationshipsList)) {
+      ts = concat(ts, ...map(relationshipsList, 'target'));
+    }
+  }
+  ts = uniqBy(ts, 'id');
```
polomani, line 619: could you refactor this a bit?
Maybe make two variables
```
let timeseriesFromAssets = []
let timeseriesFromRelationships = []
```
and do concat only once in the end.

something like:

```
ts = uniqBy([...timeseriesFromAssets, timeseriesFromRelationships], 'id')
```
you don't need `isEmpty` checks I believe

src/datasource.ts
```diff
@@ -564,11 +576,12 @@ function getDataQueryRequestType({ tab, latestValue }: QueryTarget) {
     }
   }
 }
-
 async function findAssetTimeseries(
```
polomani, line 579: rename the function to something like
```
findTsByAssetAndRelationships
```

src/datasource.ts
polomani:
```
let timeseriesFromAssets = assetQuery.includeSubTimeseries !== false ? await getTimeseries({ filter, limit }, connector) : [];
```
as simple as that?

src/datasource.ts
```diff
@@ -577,11 +590,33 @@ async function findAssetTimeseries(
...
+      limit,
+      connector
+    );
+    if (!isEmpty(relationshipsList)) {
+      ts = concat(ts, ...map(relationshipsList, 'target'));
+    }
+  }
+  ts = uniqBy(ts, 'id');
   if (ts.length === limit) {
     emitEvent(responseWarningEvent, { refId, warning: TIMESERIES_LIMIT_WARNING });
```
polomani, line 621: this limit should be checked on both arrays before you merge them. we need to find a good way of doing that

KissSpinobleZsolt: should we emit warning for each timeseries limit warning and also have the combined limit warning?

KissSpinobleZsolt: what if the length is bigger then limit? because right now we check only if equals..

polomani: sure, can be `>=` instead

muradsater:
This looks good now in my local testing. Are you happy @polomani ?

src/__tests__/datasource.test.ts
```diff
@@ -199,11 +200,15 @@ describe('Datasource Query', () => {
...
 
+  //
   describe('Given "Select Timeseries from Asset" queries', () => {
     let result;
     const assetQuery = {
       target: '789',
       includeSubtrees: false,
+      withRelationships: false,
+      includeSubTimeseries: true,
+      relationshipsQuery: defaultRelationshipsQuery,
```
polomani, line 211: this test is supposed work with the older JSON too revert these changes.

polomani (approved):
LGTM

PR #181, opened 2022-05-20
Events advanced filter support

KissSpinobleZsolt:


polomani:
Currently, there are 3 ways of querying the data in the plugin:

- GUI-based (time series, time series under asset, relationships)
- PromQl-like query language (custom query, events)
- GraphQl

This PR aims to introduce a 4th way of doing it by exposing JSON APIs directly.
We're not exactly proud of our own query language (PromQL-ish), which was more or less due to legacy reasons, but adding another one (event if it's commonly used by dev community) feels like a slippery path.

Ideally, the plugin needs to have fewer ways of doing things. This affects the learning curve, UX, and adoption.
Have you considered creating a GUI constructor for the advanced event query?

muradsater:
> Have you considered creating a GUI constructor for the advanced event query?

It's a bit early. We just want to get this in as a preview feature to make it available for the users. Making a GUI up front is a lot of work and cost to take until we have adoption.

src/types.ts
```diff
@@ -27,6 +27,13 @@ export const TabTitles = {
   [Tab.Templates]: 'Templates',
 };
 
+const defaultEventQuery: EventQuery = {
+  expr: '',
+  columns: ['externalId', 'type', 'subtype', 'description', 'startTime', 'endTime'],
+  activeAtTimeRange: true,
+  eventQuery: ``,
```
polomani, line 34: `eventQuery.eventQuery` doesn't feel right. maybe `advancedFilter`?

src/types.ts
```diff
@@ -157,6 +158,7 @@ export interface EventQuery {
   expr: string;
   activeAtTimeRange: boolean;
   columns: string[];
+  eventQuery: string;
```
polomani, line 161:
```suggestion
  advancedFilter: string;
```

src/datasource.ts
```diff
@@ -296,22 +303,34 @@ export default class CogniteDatasource extends DataSourceApi<
...
+  async fetchEvents({ expr, eventQuery }, timeRange: EventsFilterTimeParams) {
+    let filter = [];
+    let params = {};
+    if (expr) {
+      const parsedQuery = parseQuery(expr);
+      filter = parsedQuery.filters;
+      params = parsedQuery.params;
+    }
+    const advancedFilter =
+      this.connector.isEventsAdvancedFilteringEnabled() && eventQuery.length
```
polomani, line 315:
```suggestion
      this.connector.isEventsAdvancedFilteringEnabled() && eventQuery
```
this is safer, and cleaner

src/datasource.ts
```diff
@@ -296,22 +303,34 @@ export default class CogniteDatasource extends DataSourceApi<
...
       limit: EVENTS_PAGE_LIMIT,
     };
-
     const items = await this.connector.fetchItems<CogniteEvent>({
       data,
       path: `/events/list`,
       method: HttpMethod.POST,
+      headers: {
+        'cdf-version': 'alpha',
+      },
```
polomani, line 329: this header should only appear if "advancedFilter" is used. otherwise it's unsafe to use in production

polomani (commented):
I've put some comments.
Also, I believe styling is a bit off?

Screenshot 2022-07-29 at 16 51 00

src/components/queryEditor.tsx
```diff
@@ -276,7 +277,7 @@ function AssetTab(props: SelectedProps & { datasource: CogniteDatasource }) {
         },
       });
     }
-  }, [current.externalId]);
+  }, [current]);
```
polomani, line 280: Let's revert to what was here before and add relationships logic on top:

```

  useEffect(() => {
    const relationshipsQuery = current.externalId ? {
      ...query.assetQuery.relationshipsQuery,
      sourceExternalIds: [current.externalId],
    } : null;
    onQueryChange({
      assetQuery: {
        ...query.assetQuery,
        relationshipsQuery
        target: current.value,
      },
    });
  }, [current.value, current.externalId]);
```
Otherwise, it's hard to see how it's affecting the current dashboards..

polomani (approved):

PR #196, opened 2022-09-12
Event aggregates preview

muradsater:


src/types.ts
```diff
@@ -161,6 +166,8 @@ export interface EventQuery {
   activeAtTimeRange: boolean;
   columns: string[];
   advancedFilter: string;
+  withAggregate: boolean;
+  property: string[];
```
polomani, line 170: I suggest having this as a nested object:
```suggestion
  aggregate?: {
         type: "uniqueValues" | "count",
         properties: { property: string }[]
  }
```

src/datasources/EventsDatasource.ts
```diff
@@ -0,0 +1,108 @@
...
+        return {
+          advancedFilter: advancedFilterQuery,
+          filter: { ...timeRange, ...params },
+          limit: EVENTS_PAGE_LIMIT,
+        };
+      }
+      return {
+        filter: { ...timeRange, ...params },
+        limit: EVENTS_PAGE_LIMIT,
+      };
```
polomani, line 92:
```
filter: { ...timeRange, ...params },
```
this line is common for all the cases, you need to move it out
```
const filter =   { ...timeRange, ...params }
```

polomani: also, rename the function.
data is too generic, maybe `getRequestBody`?

polomani (commented):
Refactor is good, but we need to improve the json structure, by encapsulating aggregate params into one object.

polomani:
Screenshot 2022-09-15 at 12 52 50
Also I didn't manage to get this to work, @KissSpinobleZsolt

muradsater:
Did you specify an advanced filter query? It's not supported for normal
queries since it's deprecated in the API

muradsater:
Advanced query could be
```
{
    "equals": {
        "property": ["RULE_BROKEN"],
        "value": "Anomaly"
    }
}
```

src/types.ts
```diff
@@ -166,8 +169,11 @@ export interface EventQuery {
   activeAtTimeRange: boolean;
   columns: string[];
   advancedFilter: string;
-  withAggregate: boolean;
-  property: string[];
+  aggregate: {
```
polomani, line 172: this must be an optional type

src/__tests__/datasource.test.ts
```diff
@@ -650,8 +650,11 @@ describe('Datasource Query', () => {
           activeAtTimeRange: false,
           columns: [''],
           advancedFilter: '',
-          withAggregate: false,
-          property: [],
+          aggregate: {
```
polomani, line 653: pretty sure this is not supposed to be here.
aggregate is `undefined` for all legacy dashboards

PR #199, opened 2022-10-14
feat: relationships depth limit parameter

KissSpinobleZsolt:


polomani:
Thanks for the PR, we need a bit more details here:
- description
- screenshots
- example of a query

muradsater:
image

muradsater:
This is an extension to the relationsships tab to limit the number of nodes based on a start node. You can choose the depth and it will do recursive queries to get the next level of nodes in the graph

src/components/queryEditor.tsx
```diff
@@ -311,11 +311,7 @@ export function QueryEditor(props: EditorProps) {
   }, [tab]);
 
   const tabId = (t) => {
-    if (
-      t === Tabs.Templates ||
-      t === Tabs.FlexibleDataModelling ||
-      t === Tabs.ExtractionPipelines
-    ) {
+    if (t === Tabs.Templates || t === Tabs.FlexibleDataModelling) {
```
polomani, line 314: this is something I cannot agree with. extraction pipelines must stay in preview until we gather more feedback.
it doesn't matter if it's stable API or not. in the connector it's still an experimental feature

src/datasources/RelationshipsDatasource.ts
```diff
@@ -176,4 +224,8 @@ export class RelationshipsDatasource {
       return [];
     }
   }
+  getSourceExternalIds(query: RelationshipsQuery & { refId: string }): Promise<SelectableValue[]> {
+    console.log('query', query);
```
polomani, line 228:
```suggestion
```

src/components/relationships.tsx
```diff
@@ -91,6 +157,48 @@ export const RelationshipsTab = (
           }
         />
       </Field>
+      {!queryBinder && (
+        <>
+          <Field label="Start ExtrenalId" className="relationships-select">
```
polomani, line 162:
```suggestion
          <Field label="Start ExternalId" className="relationships-select">
```
may I ask you to install spell checker?

KissSpinobleZsolt: I use one but seems still not working (under "strings" at least)

src/cdf/client.ts
```diff
@@ -1,5 +1,6 @@
 import { get, cloneDeep } from 'lodash';
 import { TableData, TimeSeries } from '@grafana/data';
+import { getTemplateSrv } from '@grafana/runtime';
```
abdullah-cognite, line 3: Unused import, probably we can safely remove this?

src/cdf/helper.ts
```diff
@@ -3,15 +3,16 @@ import _ from 'lodash';
...
-      containsAny: labels.containsAny.map(({ value }) => ({ externalId: value })),
+      containsAny: _.map(labels.containsAny, ({ value }) => ({ externalId: value })),
     },
   };
-export const filterExternalId = (sourceExternalIds) =>
+
+export const filterExternalId = (sourceExternalIds, targetTypes: boolean) =>
   !_.isEmpty(sourceExternalIds) && {
-    targetTypes: ['timeSeries'],
+    targetTypes: targetTypes ? ['timeSeries'] : undefined,
```
abdullah-cognite, line 12: The variable name `targetTypes` seems vague considering that its a `boolean`, should be better off if renamed to `isTypeTimeseries` or `isTimeseries`

src/cdf/types.ts
```diff
@@ -238,5 +239,6 @@ export interface RelationshipsFilter {
     min: number;
   };
   sourceExternalIds?: string[];
-  targetTypes?: string[];
+  targetTypes?: boolean;
```
abdullah-cognite, line 242: As my comment above, The variable name `targetTypes` seems vague considering that its a `boolean`, should be better off if renamed to something with an `is<VarName>` like `isTypeTimeseries`

src/components/relationships.tsx
```diff
@@ -47,7 +48,60 @@ export const RelationshipsTab = (
 ) => {
   const { datasource, query, onQueryChange, queryBinder } = props;
   const route = queryBinder ? `${queryBinder}.${queryTypeSelector}` : `${queryTypeSelector}`;
-
+  const [options, setOptions] = useState([]);
+  const getOptions = async () => {
```
abdullah-cognite, line 52: Recommendation: Lets rename this to `updateOptions`, since this doesnt really `get` anything

src/datasources/FlexibleDataModellingDatasource.ts
```diff
@@ -37,7 +43,7 @@ export class FlexibleDataModellingDatasource {
 
   async listFlexibleDataModelling(refId: string): Promise<FDMQueryResponse> {
     try {
-      const response = await this.connector.fetchQuery({
+      const { data } = await this.connector.fetchQuery<FDMResponse>({
```
abdullah-cognite, line 46: 👍

src/types.ts
```diff
@@ -208,8 +211,9 @@ export interface RelationshipsQuery {
     min: number;
   };
   sourceExternalIds?: string[];
-  targetTypes?: string[];
+  targetTypes: boolean;
```
abdullah-cognite, line 214: As my comment above, variable name should be updated to reflect that its a `boolean`

muradsater:
@abdullah-cognite Is it ok now? Can we merge this one?

src/datasources/TimeseriesDatasource.ts
```diff
@@ -69,35 +68,39 @@ async function findTsByAssetAndRelationships(
...
+      timeseriesFromAssets.splice(-1);
+    }
+    if (timeseriesFromRelationships.length >= limit) {
+      emitEvent(responseWarningEvent, { refId, warning: TIMESERIES_LIMIT_WARNING });
+      timeseriesFromRelationships.splice(-1);
+    }
+    return _.uniqBy([...timeseriesFromAssets, ...timeseriesFromRelationships], 'id');
+  } catch (error) {
+    handleError(error, refId);
+    return [];
```
polomani, line 95: there is some error handling later in the stack, why do we need it here?
this might then call datapoints api with an empty list, which doesn't make sense anyway. so it's better to break the logic here

src/components/relationships.tsx
```diff
@@ -47,7 +48,60 @@ export const RelationshipsTab = (
...
+    },
+    [options]
+  );
+  const eventsSubscribe = async () => {
+    const appEvents = await appEventsLoader;
+    appEvents.on(visNodeGraphPanelClickEvent, handleSelectedItem);
+  };
+  const eventsUnsubscribe = async () => {
+    const appEvents = await appEventsLoader;
+    appEvents.off(visNodeGraphPanelClickEvent, resetSource);
```
polomani, line 82: datasource is listening to a panel plugin events? is there any better way to solve this, so we don't have to couple write a plugin-specific logic?

KissSpinobleZsolt: At first it felt to be a good idea to use the grafan-event-handler to query click event from the  https://github.com/cognitedata/cognite-grafana-relationships-visualization plugin when a node was selected. We can take this out and maybe is there a "better" way, to handled the click event, as I remember it were problems to call the datasource(s) in a plugin project. I assume, in this phase is better to take this event-listener out and try other scenarios? Please make a decision how to perform forward.

muradsater:
You can take it out for now, so we get this plugin released :-)

polomani:
> You can take it out for now, so we get this plugin released :-)
agreed

PR #201, opened 2022-10-27
bug-fixed

KissSpinobleZsolt:
**Problem / User story**

Under the FDM-tab, selecting the FDM-Model and the corresponding version, by triggering the query-field the Model and Version select filed loose selected values.

**Solution**

Refactor code to eliminate bug:
 - change: ```AsyncSelect``` to ```Select``` 
 - create FDM-Model state options
 -  removed ```triggerQuery``` function
 - remove ```useCallback``` at ```updateGraphQuery``` function
 - update ```useEffect``` to set Model options
 
Refactor code to lose lodash import:
 - moved function which use lodash to utils.ts

**Affected Resource types**

FDM (tab-component)

None

No

**Have tests been updated?**

No/Future PR

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -78,7 +78,7 @@ export const FDMResponseToDropdown = (edges) => {
   });
   return { all, names, dataModelOptions: sortBy(values, ['label']) };
 };
-export const reverseSort = (list) => sortBy(list).reverse();
+export const reverseSortGet = (list, id) => sortBy(get(list, id)).reverse();
 const getNodeSelection = (selection) => {
   const { selectionSet } = selection[0];
   if (selectionSet) {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-import { isNil, omitBy, get, map, assignIn, sortBy, filter, find, uniq } from 'lodash';
+import { isNil, omitBy, get, map, assignIn, sortBy, filter, find, uniq, head } from 'lodash';
 import { TimeRange } from '@grafana/data';
 import { stringify } from 'query-string';
 import ms from 'ms';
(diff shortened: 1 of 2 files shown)


polomani:
@KissSpinobleZsolt hey, we have now introduced this PR template to the repo, however it wasn't picked here as it was merged after you opened the pr.

I know it's most likely sowhere in slack, but could you fill the template for the better correspondence?

src/utils.ts
```diff
@@ -62,3 +62,45 @@ export const getFirstSelection = (graphQuery, refId) => {
...
+    });
+    assignIn(all, { [externalId]: collector });
+    values.push({ value: externalId, label: name });
+    assignIn(names, { [externalId]: name });
+  });
+  return { all, names, dataModelOptions: sortBy(values, ['label']) };
+};
+export const reverseSortGet = (list, id) => sortBy(get(list, id)).reverse();
+const getNodeSelection = (selection) => {
+  const { selectionSet } = selection[0];
```
abdullah-cognite, line 83: we should use the lodash head function here, instead of directly accessing the first element on the array, that might throw an error.

src/utils.ts
```diff
@@ -62,3 +62,45 @@ export const getFirstSelection = (graphQuery, refId) => {
...
+    return getNodeSelection(selections);
+  }
+  return [];
+};
+export const typeNameList = (selected) =>
+  uniq(
+    filter(
+      map(getNodeSelection(selected), ({ selectionSet, name: { value } }) => {
+        if (selectionSet) {
+          const i = find(selectionSet.selections, ({ name: { value } }) => value === '__typename');
```
abdullah-cognite, line 98: what does i correspond to here? can we get a more meaningful variable name?

abdullah-cognite (approved):

PR #205, opened 2022-11-17
FDM null value bugfix / items + variable support

muradsater:
**Problem / User story**
* The plugin would fail to fetch time series in case of some null values returned in some rows
* Alternative syntax of graphql supported to pull `items` instead of `node/edge` (added in FDM lately)
* Update variables to make it a config option to return externalId or name instead of internal id which makes it more usable with FDM

**Solution**
Check for null values + additiontal parsing

**Affected Resource types**
FDM, variables

N/A

No

**Have tests been updated?**
No

**Screenshots and examples**
image


polomani:
I am starting to feel a bit less conmfortable including more features into one release.
could we move first two into current 3.0.0:
```
- The plugin would fail to fetch time series in case of some null values returned in some rows
- Alternative syntax of graphql supported to pull items instead of node/edge (added in FDM lately)
```

and then have the variable extension feature later on?

muradsater:
I'd prefer to have the variable change in the same release, since it's tightly coupled to the FDM feature. Internal ids will be very rare in an FDM instance, so taking it out will make it impossible to create my demo dashboard:
image

But you decide...

muradsater:
In docs? Or here (from ice-cream-demo)
```{
  listEquipment(filter: {Facility: {eq: "${site}"}})  {
    items {
        Quality {
        externalId
          __typename
          name
        }
        OEE_calculation {
        externalId
          __typename
          name
        }
        PlannedRuntime {
          __typename
          name
        externalId
        }
      }
    }
}
```

polomani (approved):
LGTM

abdullah-cognite (approved):

PR #208, opened 2022-12-06
fix : use event bus from @grafana/data, remove SystemJS runtime

muradsater:
**Problem / User story**
Update implementation after grafana review

### DIFF
--- a/src/appEventHandler.ts
+++ b/src/appEventHandler.ts
@@ -1,14 +1,12 @@
-import { AppEvent } from '@grafana/data';
-import { SystemJS } from '@grafana/runtime';
+import { AppEvent, EventBusSrv } from '@grafana/data';
 import { getCalculationWarnings, getLimitsWarnings, stringifyError } from './cdf/client';
 import { failedResponseEvent, responseWarningEvent } from './constants';
 import { SuccessResponse } from './types';
 
-export const appEventsLoader = SystemJS.load('app/core/app_events');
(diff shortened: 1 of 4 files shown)


src/appEventHandler.ts
```diff
@@ -1,14 +1,12 @@
-import { AppEvent } from '@grafana/data';
-import { SystemJS } from '@grafana/runtime';
+import { AppEvent, EventBusSrv } from '@grafana/data';
 import { getCalculationWarnings, getLimitsWarnings, stringifyError } from './cdf/client';
 import { failedResponseEvent, responseWarningEvent } from './constants';
 import { SuccessResponse } from './types';
 
-export const appEventsLoader = SystemJS.load('app/core/app_events');
+export const bus = new EventBusSrv();
```
abdullah-cognite, line 6: Renaming this to `busService` or `eventBusService` would` be a good idea, just `bus` feels a bit ambiguous

abdullah-cognite:
Apart from the small comment above, the PR looks good to me 🚀

polomani:
Seems to work, let's do the refactor suggested by @abdullah-cognite and merge

muradsater:
> Seems to work, let's do the refactor suggested by @abdullah-cognite and merge

Could you apply the change @abdullah-cognite ? Zsolt is not available at the moment

abdullah-cognite (approved):

PR #230, opened 2023-03-30
Bump numerous dependencies

polomani:
This PR is to fix CVEs primarily, but also update the codebase where required.

- migrate to @grafana/create-plugin from @grafana/toolkit (new recommended way for plugin dev)
- all files in `.config/*` are generated by @grafana/create-plugin

Those files are also generated by @grafana/create-plugin:
- .eslintrc
- .nvmrc
- .prettierrc.js
- docker-compose.yaml
- jest-setup.js
- jest.config.js
- tsconfig.json

Dependencies and commands in package.json are updated by the script as well.

**Actual changes:**
- added `getDataSourceWithMocks` so we can initiate data source for every describe(..), the previous approach doesn't seem to work with the new setup
- moved test_utils outside of __tests__ folder
-  updated "@grafana/data": "^9.4.7", "@grafana/runtime": "^9.4.7", "@grafana/ui": "^9.4.7"

This needs to be tested with Grafana v8 / v9

Bumped Node to v16 in GitHub actions.

### DIFF
--- a/src/components/devFeatures.tsx
+++ b/src/components/devFeatures.tsx
@@ -5,7 +5,6 @@ import React, { useEffect } from 'react';
  * This is used to hide/show experimental features
  */
 export const KonamiTracker = ({ onCheat }: { onCheat: () => void }) => {
-  
   useEffect(() => {
     const keySequence = [38, 38, 40, 40, 37, 39, 37, 39, 66, 65];
     let pointer = 0;
--- a/src/components/eventsTab.tsx
+++ b/src/components/eventsTab.tsx
@@ -222,7 +222,9 @@ const AdvancedEventFilter = (props) => {
     return jsonlint.parse(advancedFilter) ?? false;
   };
   const onChange = (advancedFilter) => {
-    if (valid(advancedFilter)) {patchEventQuery({ advancedFilter });}
+    if (valid(advancedFilter)) {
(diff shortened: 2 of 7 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
+version: '3.0'
+
+services:
+  grafana:
+    container_name: 'cognitedata-datasource'
+    build:
+      context: ./.config
+      args:
+        grafana_version: ${GRAFANA_VERSION:-9.3.8}
```
polomani, line 9: we should try this with v8.5~ and 9.X

.github/workflows/release.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       - name: Install dependencies
         run: yarn install
 
+      - name: Test plugin
+        run: yarn test:ci
+
```
polomani, line 25: we need to sign the plugin as well (change the command)

polomani: update the whole file

.github/workflows/release.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       - name: Install dependencies
```
polomani, line 20: use node 16

abdullah-cognite (approved):

PR #305, opened 2023-11-07
[POFSP-222] Fix regressions in events table and annotations query

polomani:
**Problem / User story**

> Customer reported this issue after upgrading Grafana connector to 4.0.0 on their Grafana environment.
The issue is that activeAtTime filter of Events applies always regardless of on or off for the "Active only" property. 

**Solution**

Had to add some unit tests after they were removed in 4.0.0, and found another bug where annotation query filters also were not propagated properly.
Also added hints/examples panel for the query language used in annotations editor.

Screenshot:
Screenshot 2023-11-07 at 12 10 24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,26 @@
 
 This article documents the ongoing improvements we're making to the **Cognite Data Source for Grafana**.
 
+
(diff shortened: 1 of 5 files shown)


abdullah-cognite (approved):
LGTM

PR #306, opened 2023-11-20
Remove redundant dockerfile, fix documentation

polomani:
- Dockerfile is not maintained by us anymore, instead, we use one provided by Grafana
- Move documentation outside of the main readme (so it doesn't appear on grafana.com)


abdullah-cognite (approved):
lgtm

PR #320, opened 2024-02-02
fix(deprecated-apis): remove backendSrv.datasourceRequest

polomani:
**Problem / User story**
- introduced Fetcher, a simple type with a `fetch()` method, to encapsulate an HTTP client
- since we are mocking the fetcher in the tests, I made it possible to supply a custom one into the initSources method
- resolved incompatibility issue where `@grafana/runtime@^10.3.1` type didn't play well with the ~10.2 (by just updating the dependency)
- resolved deprecation warning in which backendSrv.datasourceRequest should be replaced with backendSrv.fetch, ref https://community.grafana.com/t/how-to-migrate-from-backendsrv-datasourcerequest-to-backendsrv-fetch/58770

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -67,7 +67,7 @@ export default class CogniteDatasource extends DataSourceApi<
   timeseriesDatasource: TimeseriesDatasource;
   flexibleDataModellingDatasource: FlexibleDataModellingDatasource;
 
-  constructor(instanceSettings: DataSourceInstanceSettings<CogniteDataSourceOptions>, fetcher?: Fetcher) {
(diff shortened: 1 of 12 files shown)


src/datasource.ts
```diff
@@ -84,17 +92,22 @@ export default class CogniteDatasource extends DataSourceApi<
...
       enableTemplates,
       enableEventsAdvancedFiltering,
       enableFlexibleDataModelling,
       enableExtractionPipelines
     );
+    this.initSources(connector);
+  }
+  
+  initSources (connector: Connector) {
+    this.connector = connector;
```
lpereiracgn, line 110: Why to move the connector init to a separate method instead of keeping it in the constructor?

lpereiracgn (approved):

PR #345, opened 2024-02-12
fix: scientific notation for numbers in custom query

polomani:
**Problem / User story**
In custom queries, we observe that the decimal numbers convert to scientific notation when it is more than 7 decimal places and the query fails.

**Solution**
Use standard grammar from nearley to parse float numbers (supports scientific notation as well as a side effect).

Screenshot 2024-02-12 at 09 57 00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 This article documents the ongoing improvements we're making to the **Cognite
 Data Source for Grafana**.
 
+## 4.1.1 - February 14th, 2024
+
+### Bug fixes
+
+- Fixed an issue with the custum query where floats with more than 7 digits after decimal point where not supported
+
 ## 4.1.0 - February 7th, 2024
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Data Source for Grafana**.
 
 ### Bug fixes
 
-- Fixed an issue with the custom query where floats with more than 7 digits after the decimal point were not supported
+- Fixed an issue with the custom query where floats with more than 7 digits after the decimal point were causing syntax error
 
 ## 4.1.0 - February 7th, 2024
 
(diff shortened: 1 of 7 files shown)


src/parser/events-assets/grammar.ts
```diff
@@ -56,7 +55,7 @@ interface NearleyLexer {
   reset: (chunk: string, info: any) => void;
   next: () => NearleyToken | undefined;
   save: () => any;
-  formatError: (token: NearleyToken) => string;
+  formatError: (token: never) => string;
```
sagar-thalwar, line 58: question(non-blocking): What is the reason for changing the type to `never`?

polomani: file was generated. I assume this was changed in the newer version of nearley for some reason.. not sure

sagar-thalwar (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 This article documents the ongoing improvements we're making to the **Cognite
 Data Source for Grafana**.
 
+## 4.1.1 - February 14th, 2024
+
+### Bug fixes
+
+- Fixed an issue with the custum query where floats with more than 7 digits after decimal point where not supported
```
polomani, line 10:
```suggestion
- Fixed an issue with the custom query where floats with more than 7 digits after the decimal point were not supported
```

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 This article documents the ongoing improvements we're making to the **Cognite
 Data Source for Grafana**.
 
+## 4.1.1 - February 14th, 2024
+
+### Bug fixes
+
+- Fixed an issue with the custom query where floats with more than 7 digits after the decimal point were not supported
```
polomani, line 10:
```suggestion
- Fixed an issue with the custom query where floats with more than 7 digits after the decimal point were causing syntax error
```

PR #532, opened 2025-04-11
chore: remove deprecated gf-form-switch

polomani:
**Problem / User story**
some css styles are about to be removed: https://github.com/cognitedata/cognite-grafana-datasource/issues/384
In our case it's `gf-form-switch` which used for `Switch` component.

**Solution**
- `Switch` is replaced with `InlineSwitch` which is essencially the same.
- additionally fixed the empty `Observable` response when there no valid queries (otherwise it's failing e2e test, due to infinite loading). This bug was introduced when recently when we started using observables.

**Have tests been updated?**
- yes, added e2e test that with the switches (timeseries from asset query)
- added a fixture to wait for the panel to finish loading

**Screenshots and examples**

Screenshot 2025-04-11 at 15 04 13

### DIFF
--- a/src/components/resourceSelect.tsx
+++ b/src/components/resourceSelect.tsx
@@ -89,7 +89,7 @@ export function ResourceSelect(props: {
         defaultOptions
         value={current.value ? current : null}
         placeholder={`Search ${resourceType.toLowerCase()} by name/description`}
-        className="cognite-dropdown width-20"
+        className="cog-mr-4 width-20"
         onChange={onDropdown}
       />
       <FormField
--- a/src/components/resourceSelect.tsx
+++ b/src/components/resourceSelect.tsx
@@ -49,6 +49,9 @@ export function ResourceSelect(props: {
   };
 
   const fetchDropdownResource = async (id: IdEither) => {
+    if (('externalId' in id && !id.externalId) || ('id' in id && !id.id)) {
+      return null;
+    }
     try {
       const [res] = await fetchSingleResource(id);
       return res;
--- a/src/css/query_editor.css
+++ b/src/css/query_editor.css
@@ -10,14 +10,6 @@
   min-width: 20rem;
 }
 
-.gf-form-select-wrapper select.gf-form-input {
-  height: 2.64rem;
-}
-
-.gf-form-select-wrapper--caret-indent.gf-form-select-wrapper::after {
-  right: 0.775rem;
-}
-
 .gf-tabs-cognite.active:before,
 .gf-tabs-cognite.active:focus:before,
 .gf-tabs-cognite.active:hover:before {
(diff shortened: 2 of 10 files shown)


src/components/resourceSelect.tsx
```diff
@@ -49,6 +49,9 @@ export function ResourceSelect(props: {
   };
 
   const fetchDropdownResource = async (id: IdEither) => {
+    if (('externalId' in id && !id.externalId) || ('id' in id && !id.id)) {
+      return null;
+    }
```
polomani, line 54: happens on new (empty) panels in the editor, a bit annoying

src/css/query_editor.css
```diff
@@ -33,8 +33,7 @@ pre code,
   line-height: 2;
 }
 
-.cognite-dropdown,
-.gf-form-switch {
+.cognite-dropdown {
   margin-right: 4px;
 }
```
lpereiracgn, line 38: I see we're only using this property in `AsyncSelect => className="cognite-dropdown width-20"`
Do we really need a CSS property for setting that? Is it possible to use a helper similar to `width-{x}`?

src/parser/ts/parser.unit.test.ts
```diff
@@ -543,8 +543,8 @@ describe('parse unary "-" operator', () => {
       expect(parse(query)).toEqual(expected);
       const endTime = Date.now(); 
       const timeDiff = endTime - startTime; 
-      // in case you have to bump the limit up from 20ms, make sure you know what you are doing :)
-      expect(timeDiff).toBeLessThan(20);
+      // in case you have to bump the limit up from 50ms, make sure you know what you are doing :)
+      expect(timeDiff).toBeLessThan(50);
```
lpereiracgn, line 547: Why the bump? 👀

tests/queryEditor.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        return false;
+      }
+    }
+  )).toBeOK();
+
+  await expect.poll(async () => {
+    const buttonTexts = await panelEditPage.panel.locator.getByRole('button', { name: /59.9139-10.7522.*/ }).allInnerTexts();
+    return buttonTexts.sort();
+  }, { timeout: 10000 }).toEqual(expectedTs);
+});
```
lpereiracgn, line 73: Why to use `expect.poll` here?

src/parser/ts/parser.unit.test.ts
polomani: flaky, 20 vs 50 should not be a big difference for the user

tests/queryEditor.spec.ts
polomani: Need to sort them before using `expect`, because of this we lose native timeout behaviour of expect..
otherwise it'll try to find buttons and if they are not there it'll fail immediately.
do you know any other way?

src/css/query_editor.css
polomani: I renamed this to `cog-mr-4`, since `mr-4` doesn't work (I guess there is no tailwind under the hood)

polomani: and removed unused classes

polomani: renamed it to `cog-mr-4` just to make it obvious it's a helper (there is no standard one there, unfortunately)

lpereiracgn (approved):

PR #533, opened 2025-04-23
chore: remove deprecated data structure interface

polomani:
**Problem / User story**
- `MutableDataFrame` is deprecated, `DataFrame` should be used instead

**Solution**
- remove the usage of `MutableDataFrame`

**Affected Resource types**
- templates
- relationships

No

**Have tests been updated?**
- added e2e tests for the relationships panel

**Screenshots and examples**
Screenshot 2025-04-23 at 14 19 03


src/__tests__/__snapshots__/relationships.test.ts.snap
```diff
@@ -80,6 +74,7 @@ exports[`Relationships createRelationshipsNode adds value when relationshipsList
         ],
       },
     ],
+    "length": 2,
```
abdullah-cognite, line 77: question: what is the length here?

polomani: number of items in the columns. if you look up each has 2
```
        "values": [
          "test-source",
          "test-target",
        ],
```

src/datasources/RelationshipsDatasource.ts
```diff
@@ -48,28 +50,30 @@ export const createRelationshipsNode = (relationshipsList, refId): Relationships
...
+      values: []
     })),
     meta: {
       preferredVisualisationType: 'nodeGraph',
     },
     refId,
-  });
-  const nodes = new MutableDataFrame({
+  };
+  const nodes: Partial<DataFrame> = {
```
abdullah-cognite, line 65: any reasons behind using `Partial` or was that how Grafana suggested to use the `Dataframe`?

polomani: need a type that is not exactly Dataframe yet, but could be converted to Dataframe. 
We can define an interface ourselves, but this seemsed to be simpler

abdullah-cognite (approved):

PR #672, opened 2026-08-31
Extract shared view selector and instance search components

evertoncolling:
**Problem / User story**

The view selector and instance search were hand-rolled per tab, which had accumulated real problems:
- Five copies of the "fetch views → map to options → match the selected triple → render a Select" block, with three copies of the same option formatter and three different conventions for the selected value.
- Both instance searches fired a request on every keystroke, with no debounce.
- The two searches encoded instance references as `space:externalId`, which mis-parses the externalIds that legitimately contain colons (`asset:equip:iaa_met_34es7512`).
- Search sent no `operator`, so the API default of `OR` applied: every extra word the user typed *widened* the result set instead of narrowing it.
- Error handling diverged — the Time Series tab had an Alert for view-load failures that was unreachable, while the other call sites only logged to the console.
- The Activities tab's Instance View lived only in component state, so a saved dashboard restored the selected instances but not the view they were searched in.

The upcoming Records tab needs the same two widgets, so this is the moment to share them rather than add a sixth copy.

**Solution**

Two new components under `src/components/common/`, each following the same shape: a props-only component plus exported pure helpers and (for views) a data hook, so tests cover the logic without a DOM and callers with bespoke rendering can reuse the parts.

- `ViewPicker` — lists the views implementing a `cdf_cdm` container. Exports `useContainerViews`, `viewLabel`, `toViewOptions`, and `encodeViewRef`/`parseViewRef`. Replaces all four hand-rolled selectors in the two tabs. `allVersions: true` is kept and now documented: without a separate version picker, the dropdown is the only way to reach an older view version.
- `InstancePicker` — searches the instances of a view. 300 ms debounce, JSON-encoded references, `operator: 'AND'`, and a unified `onError` that the Time Series tab now surfaces in its Alert.

Also here: `fetchCogniteTimeSeriesViews` and `fetchCogniteActivityViews` were line-for-line copies of the already-generic container fetcher, so they collapse into one-line wrappers (about −90 lines). The inspect request no longer asks for `totalInvolvedViewCount`, which we never read.

The Records tab will consume `useContainerViews` and the mappers rather than `<ViewPicker>` itself — its options carry stream metadata and its `onChange` migrates the query — which is why the hook and mappers are exported individually.

The seven commits are ordered dependencies-first and each builds and tests green on its own, so the branch bisects cleanly. Reviewing commit by commit is easier than reading the squashed diff.

**Affected Resource types**

CogniteTimeSeries tab (view selector, activity-overlay view selector, time series search) and Activities tab (activity view, instance view, instance search). DMS views and instances via `/models/containers/inspect` and `/models/instances/search`. No change to Data Models, Events, Assets, or any legacy tab.

One addition: `cogniteActivityTabQuery.instanceView?: { space, externalId, version }`. Optional and backward compatible — queries saved without it fall back to the `cdf_cdm` standard view, which is the default the tab always applied, so existing dashboards behave exactly as before. Nothing removed or renamed; no breaking changes.

Worth clarifying for review: the instance search's move to JSON-encoded references changes only the in-Select option value. The persisted fields are untouched — `cogniteTimeSeries.instanceId` is still `{space, externalId}` and `assetInstances` is still `Array<{space, externalId, name}>`. No dashboard migration needed.

Yes — `DEVELOPMENT.md` documents the `.env` bootstrap and the new persistence behaviour, including how to reset to a clean provisioned state.

**Have tests been updated?**

Yes. New unit tests for both components' pure helpers (option mapping, reference encoding/round-tripping, search-request construction including the `AND` operator and limit clamping) and for `useContainerViews` (load, refetch on container change, error path). The existing container-inspect request assertion was updated for the trimmed request body. Full suite: 410 tests, 25 suites, green; lint, typecheck and build also green.

**Notable behaviour changes**

1. Instance search is debounced at 300 ms instead of one request per keystroke.
2. Search sends `operator: "AND"`, so multi-word queries narrow rather than widen. Result sets will differ from before for multi-word searches — this is the intended fix.
3. The Activities tab now persists the selected Instance View, so it survives a dashboard save and reload.
4. View-load failures in the Time Series tab now reach the user through the existing Alert rather than only the console.

**Screenshots and examples**

No visual changes.


src/components/common/ViewPicker.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [views, setViews] = useState<InvolvedView[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [loaded, setLoaded] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    let cancelled = false;
+    setLoading(true);
+    setLoaded(false);
+    setError(null);
```
muradsater, line 67: **Must fix:** this effect never clears `views` when `container` changes.

On resource-type change (Activities: Asset → Equipment) the dropdown still lists the previous container’s views and they remain selectable while the next fetch is in flight. The parent already clears `instanceView` and candidates, so a user can persist a CogniteAsset view while `resourceType` is already CogniteEquipment.

Fix: `setViews([])` here (or remount `ViewPicker` with `key={container}`).

src/components/common/ViewPicker.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  useEffect(() => {
+    if (loaded) {
+      onViewsLoadedRef.current?.(views);
+    }
+  }, [loaded, views]);
+
+  useEffect(() => {
+    if (error) {
+      (onErrorRef.current ?? console.warn)(error);
```
muradsater, line 149: This only notifies when `error` is truthy, so a later successful load never clears the parent Alert.

Old `loadViews` called `setError(null)` on success; that path is gone. Please call `onError(null)` (or equivalent) on success as well, and type `onError` as `(message: string | null) => void`.

src/components/common/InstancePicker.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const search = useCallback(
+    async (query: string): Promise<InstanceOption[]> => {
+      if (!view?.space || !view?.externalId || !view?.version) {
+        return [];
+      }
+      try {
+        const request = buildInstanceSearchRequest(view, query, filter, limit);
+        return toInstanceOptions(await searchDMSInstances(connector, request), view);
+      } catch (error) {
+        onError?.(`Search failed: ${stringifyError(error)}`);
```
muradsater: Same sticky-error issue as `ViewPicker`: `onError` is only called in `catch`. A failed search leaves the Time Series Alert up even after a later successful search (or a view change).

Clear on success too, e.g. `onError(null)` after a good response.

src/components/cogniteActivityTab.tsx
```diff
@@ -298,80 +234,22 @@ export const CogniteActivityTab: React.FC<CogniteActivityTabProps> = ({
...
+    name: a.name,
   }));
 
-  const handleActivityViewChange = (selected: SelectableValue<InvolvedView>) => {
-    if (selected?.value) {
-      const v = selected.value;
+  const handleActivityViewsLoaded = (views: InvolvedView[]) => {
+    if (views.length === 1) {
+      const v = views[0];
       onQueryChange({
```
muradsater: `handleActivityViewChange` calls `loadViewProperties`; this auto-select path does not.

If the project has exactly one CogniteActivity view and it is not the mount-time `cdf_cdm` / `v1` default, columns/sort stay on the wrong property list. Please call `loadViewProperties(v)` here (and skip `onQueryChange` when the triple is already saved, so opening the editor does not dirty the dashboard).

muradsater (changes requested):
## Review

Nice extraction. JSON instance refs, `operator: AND`, debounce, and throwing `fetchContainerViews` so the Time Series Alert can actually show view-load failures are all real fixes. Helper tests look good.

One correctness bug in `useContainerViews` should be fixed before merge: stale options after a container change.

### Must fix
- `useContainerViews` does not clear `views` when `container` changes (see inline).

### Should fix
- Errors are never cleared on success (see inline).
- Auto-select of a single activity view does not reload properties (see inline).
- The PR description claims `useContainerViews` tests (load, refetch on container change, error path). What landed is fetcher + pure mapper tests. That gap is exactly what would have caught the stale-options bug.

### Nits
- Debounce cancel only covers the 300ms wait. After `search()` is in flight, a newer keystroke does not abort the old request. react-select usually drops stale replies; a generation counter would make that explicit.
- `search` depends on the `view` and `filter` objects. Callers pass new literals every render, so `loadOptions` is recreated every time. Depend on `view.space` / `view.externalId` / `view.version` (and `instanceSpace`) instead.
- Grafana persistence (`./grafana-data/`) is useful DX but unrelated to the pickers. Don’t copy that folder around — `grafana.db` can hold session/datasource secrets.
- `handleActivityViewsLoaded` always `onQueryChange`s when `views.length === 1`, even if the triple is already saved. That can dirty a dashboard on open; skip the write when nothing changed.

src/components/common/ViewPicker.tsx
evertoncolling: Addressed in [8e51c8e](https://github.com/cognitedata/cognite-grafana-datasource/commit/8e51c8e92af51913e83b6b98764dda7dfb9a1946) — the hook now empties `views` the moment `container` changes, so the previous container's views are never selectable while the next fetch is in flight. [c837f85](https://github.com/cognitedata/cognite-grafana-datasource/commit/c837f856f6bee23932da0035b0ea95eeaf61ffca) adds the regression test (verified to fail against the pre-fix hook).

src/components/common/ViewPicker.tsx
evertoncolling: Addressed in [8e51c8e](https://github.com/cognitedata/cognite-grafana-datasource/commit/8e51c8e92af51913e83b6b98764dda7dfb9a1946) — `onError` is now `(message: string | null) => void` and fires with `null` once a later load succeeds, restoring the old `setError(null)` behaviour.

src/components/common/InstancePicker.tsx
evertoncolling: Addressed in [8e51c8e](https://github.com/cognitedata/cognite-grafana-datasource/commit/8e51c8e92af51913e83b6b98764dda7dfb9a1946) — same fix here: `onError?.(null)` after a successful search. Note the Time Series tab shares one error state across both pickers, so a success in either clears the Alert — same semantics master had.

src/components/cogniteActivityTab.tsx
evertoncolling: Addressed in [9d8618c](https://github.com/cognitedata/cognite-grafana-datasource/commit/9d8618c0d8dd04bc0369b4816e102d2814405a24) — the auto-select path now calls `loadViewProperties(v)` and returns early when the triple is already saved, so opening the editor no longer dirties the dashboard. For context, both quirks exist on master (the refactor had preserved them to stay behaviour-neutral) — agreed they were worth fixing here.

evertoncolling:
Thanks for the review — all points addressed (inline replies on each thread). The rest:

- **Test-claim gap**: you were right, the description was stale. [c837f85](https://github.com/cognitedata/cognite-grafana-datasource/commit/c837f856f6bee23932da0035b0ea95eeaf61ffca) restores full `useContainerViews` coverage (load, refetch-clears-stale-views, error path, error→success) after adding the missing `@testing-library/dom` peer dep. Description updated to match.
- **Debounce / in-flight requests**: [2c77c2c](https://github.com/cognitedata/cognite-grafana-datasource/commit/2c77c2cdc5bb05926f90038ac393bbe7c4a9bd36) adds a generation counter so a superseded search resolves empty instead of relying on react-select dropping stale promises.
- **Callback deps**: same commit — `search` now depends on the primitive view fields plus a serialized filter.
- **grafana-data**: fair scope point — reverted here in [ebcb1c0](https://github.com/cognitedata/cognite-grafana-datasource/commit/ebcb1c07c9f5e7fd0c5bda9d9fd734fc920d7482) and split into its own PR, which also adds a DEVELOPMENT.md warning that `grafana.db` holds datasource credentials and session cookies.

On CI: `e2e grafana-enterprise@13.2.0` isn't failing on assertions — it hangs and gets cancelled at the 20-min timeout, identically on the last five runs including master runs predating this branch (32840811670, 32838724779, 32828067627). Pre-existing; worth its own issue.

grafana-data/storage/storage.json
```diff
@@ -0,0 +1,5 @@
+{
+  "allowUnsanitizedSvgUpload": false,
+  "addDevEnv": true,
+  "roots": null
```
muradsater: **Must fix:** this directory should not be in the PR. `grafana-data/grafana.db` (~1.5 MB) was added in the same commit ([8e51c8e](https://github.com/cognitedata/cognite-grafana-datasource/commit/8e51c8e92af51913e83b6b98764dda7dfb9a1946)) after `/grafana-data/` left `.gitignore`.

Remove `grafana-data/` from the branch. That SQLite file can contain datasource credentials and session cookies; if it did, rotate them.

muradsater (changes requested):
## Re-review

Previous comments are addressed: stale views are cleared, `onError(null)` clears the Alert, auto-select reloads properties and skips a no-op write, hook tests cover the container-change case, debounce uses a generation counter, and Grafana persistence was split out of the picker work.

One new blocker.

### Must fix

`grafana-data/grafana.db` (~1.5 MB) and `grafana-data/storage/storage.json` landed in [8e51c8e](https://github.com/cognitedata/cognite-grafana-datasource/commit/8e51c8e92af51913e83b6b98764dda7dfb9a1946) after the persistence revert dropped `/grafana-data/` from `.gitignore`. `grafana.db` is a live Grafana SQLite file and can hold datasource credentials and session cookies.

Please drop both files from the branch (and from git history on this branch if you rewrite it before merge). Do not commit `grafana-data/`. After it is gone, check whether that DB had real secrets; if it did, rotate them — the blob is already on GitHub.

### Nits

- `.npmrc` pinning `registry.yarnpkg.com` is unrelated to the pickers (it came in with the hook-test peer dep). Fine if intentional for this repo; otherwise it belongs in a separate change.
- Increment the InstancePicker generation counter on unmount as well, so an in-flight search cannot call `onError` after the tab is gone. Optional.

src/components/common/ViewPicker.tsx
muradsater: Looks good — `setViews([])` plus the hook test for the in-flight window covers it.

src/components/common/ViewPicker.tsx
muradsater: Looks good — `onError(null)` on success is what was needed.

src/components/common/InstancePicker.tsx
muradsater: Looks good — clearing on a successful search, and dropping superseded in-flight replies, is enough here.

src/components/cogniteActivityTab.tsx
muradsater: Looks good — `loadViewProperties` on auto-select and skipping the write when the triple is already saved.

grafana-data/storage/storage.json
evertoncolling: Good catch, thank you — that was an accidental `git add -A` sweep after the revert removed the `.gitignore` entry. Fixed properly:

- History rewritten and force-pushed: the blob is gone from every commit on the branch (verified per-commit), not just the tip. `.gitignore` now defensively ignores `/grafana-data/` on this branch too ([e8c4875](https://github.com/cognitedata/cognite-grafana-datasource/commit/e8c4875d50c0db5cdf18c1d91057580a28a822ab)).
- The credentials in that db are being rotated, since dangling commits stay fetchable by SHA until GC.

Rewritten SHAs for the earlier fix references: 8e51c8e → [c1b6234](https://github.com/cognitedata/cognite-grafana-datasource/commit/c1b6234f06ba0d2ece760212b4cfa958019fc8c0), 9d8618c → [2f723d0](https://github.com/cognitedata/cognite-grafana-datasource/commit/2f723d0f3fa0f6b77332f2480faf64e463acc43d), 2c77c2c → [6b35ee7](https://github.com/cognitedata/cognite-grafana-datasource/commit/6b35ee73e091196e2941868edd2f44ab130d23da), c837f85 → [2f642ff](https://github.com/cognitedata/cognite-grafana-datasource/commit/2f642ff9bddbe0501e388b19502baf727ac8608d).

muradsater (approved):

PR #675, opened 2026-09-02
Refactor feature flags into named options and one toggle row

evertoncolling:
**Problem / User story**

Feature flags reached the `Connector` as fourteen positional booleans, which mis-wire silently the moment one is inserted in the middle, and the same fourteen `key = FEATURE_DEFAULTS.key` defaults were written out again in the datasource constructor and in the config editor. The config editor rendered each toggle by hand: fourteen copies of the id, tooltip and label-width block, in which one row (the OAuth client credentials switch) had already lost its `id`, leaving its label pointing at nothing. `masterToggleHandler` in `configEditorUtils` had no caller. The e2e "save and persist" test claimed to restore the provisioned fixture but saved its inverse (CDM on, asset-centric off), and never asserted the one legacy flag whose default is off.

**Solution**

- `Connector` takes a `ConnectorOptions` object (`oauthPassThru`, `oauthClientCredentials`, plus the feature flags by name). A `flag()` accessor coerces an unset flag to `false`, matching the old `&&`-of-optionals semantics.
- `resolveFeatureFlags(jsonData)` fills in defaults once, from the `FEATURE_DEFAULTS` key list, and both the datasource and the config editor use it. Adding a flag now reaches every consumer without touching an argument list.
- `FeatureToggleRow` renders one switch with its label, tooltip and optional badge. Feature rows look their tooltip up by flag in `FEATURE_TOOLTIPS` (typed `Record<FeatureKey, string>`, so a flag without an explanation is a compile error); the two OAuth rows pass their own tooltip and label width. That restores the missing `id` structurally.
- `isLegacyDataModelFeaturesEnabled()` is exposed on its own: asset-centric variable queries hit `/assets/list`, which no sub-flag covers, so the master switch is the only meaningful gate for them. Consumed by the variable editor in a later PR.
- Dead code removed: `masterToggleHandler` and its tests, and the never-read `featureFlags` map on `CogniteDataSourceOptions`.
- e2e: the persistence test restores the fixture (asset-centric on) instead of turning both masters on in sequence, the mutual exclusion of the masters gets its own non-saving test, and re-enabling the legacy master asserts `enableEventsAdvancedFiltering` stays off.

**Affected Resource types**

None at runtime. Datasource settings UI (Connection and Features tabs) is rebuilt on the shared row but renders the same controls with the same ids. Three small settings-UI changes go beyond a pure refactor:

- Re-enabling either master now restores its sub-flags to their `FEATURE_DEFAULTS` values. The CDM master previously forced every sub-flag on; since every CDM sub-flag defaults on, the result is the same today.
- Tooltips say "GraphQL tab" where they said "Data Models tab" before. The query editor tab is labelled GraphQL, so the old copy was stale.
- Feature toggle labels are two units wider (16, was 14) so the longest labels fit beside their tooltip icon on one line. Connection tab labels keep 14.

None. The `jsonData` keys and their defaults are unchanged; only how they are read is different.

No — not applicable.

**Have tests been updated?**

Yes. `connector.featureFlags.test.ts` is rewritten as a table over every CDM and legacy feature × all four master/flag combinations (broader than before, at a fifth of the length), with cases for the standalone legacy master and for omitted flags reading as `false`. New `featureDefaults.test.ts` covers `resolveFeatureFlags`. `test_utils` derives its all-on connector from the flag list. e2e `featureFlags.spec.ts` as above. `configEditor.spec.ts` now locates the OAuth client credentials switch by its id and toggles it with `dispatchEvent` (a `switch` role only exists on Grafana 12+): once the switch carries the id its label points at, `getByLabel` matches two elements (the checkbox and the switch's inner label), which strict mode rejects. Verified locally against Grafana 13.2: config editor and feature flag specs green.

**Screenshots and examples**

N/A.

### DIFF
--- a/src/__tests__/featureDefaults.test.ts
+++ b/src/__tests__/featureDefaults.test.ts
@@ -1,0 +1,18 @@
+import { FEATURE_DEFAULTS, resolveFeatureFlags } from '../featureDefaults';
+
+describe('resolveFeatureFlags', () => {
+  it('fills in every flag a datasource leaves unset', () => {
+    expect(resolveFeatureFlags({})).toEqual(FEATURE_DEFAULTS);
+    expect(resolveFeatureFlags()).toEqual(FEATURE_DEFAULTS);
+  });
+
+  it('keeps a flag that is set, including one set off', () => {
+    const flags = resolveFeatureFlags({
+      enableCoreDataModelFeatures: false,
+      enableEventsAdvancedFiltering: true,
+    });
+    expect(flags.enableCoreDataModelFeatures).toBe(false);
+    expect(flags.enableEventsAdvancedFiltering).toBe(true);
+    expect(flags.enableLegacyDataModelFeatures).toBe(FEATURE_DEFAULTS.enableLegacyDataModelFeatures);
+  });
+});
--- a/src/__tests__/featureDefaults.test.ts
+++ b/src/__tests__/featureDefaults.test.ts
@@ -6,6 +6,11 @@ describe('resolveFeatureFlags', () => {
     expect(resolveFeatureFlags()).toEqual(FEATURE_DEFAULTS);
   });
 
+  it('treats an explicit undefined like an unset flag', () => {
+    const flags = resolveFeatureFlags({ enableCogniteTimeSeries: undefined });
+    expect(flags.enableCogniteTimeSeries).toBe(FEATURE_DEFAULTS.enableCogniteTimeSeries);
+  });
+
   it('keeps a flag that is set, including one set off', () => {
     const flags = resolveFeatureFlags({
       enableCoreDataModelFeatures: false,
--- a/src/components/FeatureToggleRow.tsx
+++ b/src/components/FeatureToggleRow.tsx
@@ -1,0 +1,55 @@
+import React, { ChangeEvent } from "react";
+import { InlineFieldRow, InlineFormLabel, InlineSwitch } from "@grafana/ui";
+import { FeatureKey } from "../featureDefaults";
+import { FEATURE_TOOLTIPS } from "./featureTooltips";
+
+/** Shared by every toggle so the switches stay in one column. */
+const FEATURE_LABEL_WIDTH = 16;
+
+type FeatureToggleRowProps = {
+  /** Doubles as the DOM id; the e2e specs locate switches by it. */
+  id: string;
+  label: string;
+  value: boolean;
+  onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+  labelWidth?: number;
+  /** Rendered after the label, e.g. a Beta chip. */
+  badge?: React.ReactNode;
+} & (
+  /** A feature flag, explained by its entry in FEATURE_TOOLTIPS. */
+  | { feature: FeatureKey; tooltip?: undefined }
+  /** Any other switch, e.g. an authentication option, with its own explanation. */
+  | { feature?: undefined; tooltip: string }
+);
+
+/** One switch with its label, explanation and optional chip. */
+export const FeatureToggleRow = ({
+  id,
+  feature,
+  tooltip,
+  label,
+  value,
+  onChange,
+  labelWidth = FEATURE_LABEL_WIDTH,
+  badge,
+}: FeatureToggleRowProps) => (
+  <InlineFieldRow style={{ marginBottom: "4px" }}>
+    <InlineFormLabel
+      htmlFor={id}
+      tooltip={tooltip ?? FEATURE_TOOLTIPS[feature]}
+      width={labelWidth}
+    >
+      {badge ? (
+        <span
+          style={{ display: "flex", alignItems: "center", gap: "6px" }}
+        >
+          {label}
+          {badge}
+        </span>
+      ) : (
+        label
+      )}
+    </InlineFormLabel>
+    <InlineSwitch id={id} label={label} value={value} onChange={onChange} />
+  </InlineFieldRow>
+);
--- a/src/components/FeatureToggleRow.tsx
+++ b/src/components/FeatureToggleRow.tsx
@@ -3,7 +3,11 @@ import { InlineFieldRow, InlineFormLabel, InlineSwitch } from "@grafana/ui";
 import { FeatureKey } from "../featureDefaults";
 import { FEATURE_TOOLTIPS } from "./featureTooltips";
 
-/** Shared by every toggle so the switches stay in one column. */
+/**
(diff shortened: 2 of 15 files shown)


src/components/configEditor.tsx
```diff
@@ -172,11 +138,7 @@ export function ConfigEditor(props: ConfigEditorProps) {
       [masterKey]: isEnabled,
     };
     dependentKeys.forEach((key) => {
-      patch[key] = isEnabled
-        ? (masterKey === "enableCoreDataModelFeatures"
-          ? true
-          : FEATURE_DEFAULTS[key])
-        : false;
+      patch[key] = isEnabled ? FEATURE_DEFAULTS[key] : false;
```
muradsater, line 144: CDM master toggle now uses `FEATURE_DEFAULTS[key]` instead of hardcoding `true` for sub-flags when enabling. With current defaults this is equivalent, but worth calling out explicitly in the PR description (it currently says "no behaviour change"). If a CDM sub-flag default ever becomes `false`, re-enabling the master would no longer force it on — which seems like the right behavior, just a subtle unification.

src/components/featureTooltips.ts
```diff
@@ -0,0 +1,39 @@
...
+    `Master toggle for Core Data Model (CDM) features. When disabled, all CDM features will be hidden. Enabling this will disable asset-centric features.`,
+  enableLegacyDataModelFeatures:
+    `Master toggle for asset-centric (legacy) features. When disabled, all asset-centric features will be hidden. Enabling this will disable CDM features.`,
+
+  enableCogniteTimeSeries:
+    `Enable the Time Series tab to browse and select time series instances from the Core Data Model (CogniteTimeSeries type).`,
+  enableCogniteActivities:
+    `Enable the Activities tab to query CogniteActivity instances from the Core Data Model.`,
+  enableFlexibleDataModelling:
+    `Enable the GraphQL tab to query custom data models in CDF using GraphQL. Supports listing, searching, and aggregating data model instances.`,
```
muradsater, line 20: User-visible copy change: tooltips now say "GraphQL tab" where they previously said "Data Models tab" (`enableFlexibleDataModelling`, `enableRelationships`). Likely intentional (UI label is already "GraphQL"), but flagging since this isn't a pure refactor — confirm with product/content if needed.

src/components/FeatureToggleRow.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { ChangeEvent } from "react";
+import { InlineFieldRow, InlineFormLabel, InlineSwitch } from "@grafana/ui";
+import { FeatureKey } from "../featureDefaults";
+import { FEATURE_TOOLTIPS } from "./featureTooltips";
+
+/** Shared by every toggle so the switches stay in one column. */
+const FEATURE_LABEL_WIDTH = 16;
```
muradsater, line 11: Minor UI tweak: feature toggle label width went from 14 to 16 (connection-tab toggles still use 14). Fine if intentional; wasn't mentioned in the PR description.

src/connector.ts
```diff
@@ -154,53 +152,72 @@ export class Connector {
...
+    return this.flag("enableCoreDataModelFeatures") &&
+      this.flag("enableFlexibleDataModelling");
   }
 
   // Legacy data model features
+  /**
+   * The master switch on its own. Asset-centric variable queries hit /assets/list,
+   * which no sub-flag covers, so the master is the only meaningful gate for them.
+   */
+  isLegacyDataModelFeaturesEnabled() {
```
muradsater, line 183: `isLegacyDataModelFeaturesEnabled()` is added without a symmetric `isCoreDataModelFeaturesEnabled()`. The PR description explains why (asset variable queries). Consider a one-line comment here that CDM has no equivalent standalone gate, so future readers don't add a symmetric accessor by mistake.

src/__tests__/featureDefaults.test.ts
```diff
@@ -0,0 +1,18 @@
+import { FEATURE_DEFAULTS, resolveFeatureFlags } from '../featureDefaults';
```
muradsater, line 1: Optional: add a case for `resolveFeatureFlags({ enableCogniteTimeSeries: undefined })` to document that explicit `undefined` falls through to the default (same as omitted).

muradsater (approved):
## Summary

Solid refactor: fixes real issues (positional boolean args, duplicated defaults, missing OAuth switch `id`, broken e2e teardown), cuts ~600 lines, and broadens test coverage. No change to the datasource `jsonData` schema — existing dashboards and saved datasource configs should remain compatible.

**Verdict: Approve with minor notes below.**

### Highlights
- `ConnectorOptions` + `flag()` removes silent mis-wiring risk as the flag list grows.
- `resolveFeatureFlags()` is a single source of truth for defaults in both datasource and config editor.
- Exhaustive `FEATURE_TOOLTIPS: Record<FeatureKey, string>` catches missing tooltips at compile time.
- Table-driven `connector.featureFlags.test.ts` and fixed e2e teardown are meaningful improvements.
- CI green across Grafana 10–13.

muradsater (approved):

src/components/configEditor.tsx
evertoncolling: Both masters now follow one rule: re-enabling restores sub-flags to their defaults. Same result today since every CDM sub-flag defaults on. Added a comment at the call site in a713c1fc5dbba6fc8e2c6bcdec5b5592318f6f1c and noted it in the description.

src/components/featureTooltips.ts
evertoncolling: The query editor tab is labelled "GraphQL", so "Data Models tab" was stale. Fixed the last remaining reference in featureDefaults.ts in a7e7b6a97bd9abb2fa3fad0bba5dbeb335911654 and called it out in the description.

src/components/FeatureToggleRow.tsx
evertoncolling: Documented on the constant in 8c1ad1c15e8456a635e66daaa68477625de8a867 and in the description.

src/connector.ts
evertoncolling: Added to the doc comment in 16201909310d4617538990a4f5e27e661f235d1f: every CDM code path is gated by a sub-flag, so there is deliberately no isCoreDataModelFeaturesEnabled().

src/__tests__/featureDefaults.test.ts
evertoncolling: Added a case for an explicit undefined in 6665cec5475485b9a878097ecf9ff0312fa6dc8b.

PR #676, opened 2026-09-09
Add @grafana/plugin-ui and the jest setup the shared GraphQL editor needs

evertoncolling:
**Problem / User story**

The shared GraphQL editor (coming in a later PR of this stack) is built on the `EditorRow` / `EditorField` primitives from `@grafana/plugin-ui`, the same layout Grafana's own datasource editors use. That package is not a dependency yet, and its transitive tree needs two things the jest scaffolding does not provide: an ESM transform for the ESM-only modules it and the GraphQL language service ship, and `IntersectionObserver` / `ResizeObserver` stubs, without which Grafana's `Select` throws instead of opening its menu under jsdom.

**Solution**

- Add `@grafana/plugin-ui@0.16.1` (pinned). Nothing imports it yet, so this PR has no user-visible change. Its peer dependencies require `@grafana/data` and `@grafana/ui` >= 10.4.7; `grafanaDependency` in `plugin.json` stays at `>=10.0.0` here and moves to `>=10.4.7` in the editor PR, together with the first import and a **Breaking changes** changelog entry.
- `renovate.json`: cap `@grafana/plugin-ui` at `<0.17.0`. From 0.17.0 it peers `@grafana/*` at `^13.0.0`, and `:automergeMinor` would otherwise automerge that bump.
- `jest.config.js`: extend `transformIgnorePatterns` with `@grafana/plugin-ui` (ESM-only nested `uuid@14`) and `vscode-languageserver-types` (ESM-only, pulled in by `graphql-language-service` for query autocomplete; jsdom resolves its `browser` export condition to the ESM build).
- `jest-setup.js`: stub `IntersectionObserver` and `ResizeObserver` when absent. The scaffolded `.config/jest-setup` is regenerated by create-plugin, so the extensions live in the repo-owned file.
- `.yarnrc` with `ignore-engines true`: `@grafana/plugin-ui` declares `engines.yarn >= 4.16` and `node >= 24`, which yarn 1 would otherwise refuse to install. The reason is in the file and in `DEVELOPMENT.md`.

The lockfile additions (about 175 lines) all trace to `@grafana/plugin-ui`'s tree (`@react-awesome-query-builder`, `sql-formatter`, `react-calendar`, `i18next`, `react-redux`, `immutable`, `uuid@14`, `moment`, and their dependencies) and resolve to `registry.yarnpkg.com` like the rest of the file.

**Affected Resource types**

None. No runtime code changes.

None.

Yes — `DEVELOPMENT.md` notes the yarn 1 engine override. No changelog entry: nothing user-visible changes until the editor PR.

**Have tests been updated?**

No new tests; the jest configuration changes are exercised by the editor tests in the follow-up PR. Verified locally: `yarn install --frozen-lockfile`, typecheck, lint, the unit suite (420 tests) and the production build (with `>=10.0.0` in `dist/plugin.json`) all pass.

**Screenshots and examples**

N/A.


src/plugin.json
```diff
@@ -35,7 +35,7 @@
   },
 
   "dependencies": {
-    "grafanaDependency": ">=10.0.0",
+    "grafanaDependency": ">=10.4.7",
```
muradsater: This is the only user-visible change, and `@grafana/plugin-ui` is not imported yet. If master is released after this merges and before the GraphQL editor PR, we drop Grafana 10.0.0–10.4.6 for a dependency nobody uses.

Either land the consumer PR immediately after this, or keep `grafanaDependency` at `>=10.0.0` until something actually imports `EditorRow` / `EditorField`. Yarn 1 will only warn on the unsatisfied peers in the meantime.

Grafana’s plugin-ui docs also suggest considering a plugin major when raising this floor so catalog/CLI users on older Grafana do not pick up an incompatible zip. For 10.0 → 10.4.7 I would not block on 5.0.0, but it is worth a conscious call.

package.json
```diff
@@ -127,6 +127,7 @@
   "dependencies": {
     "@emotion/css": "^11.1.3",
     "@grafana/data": "^11.3.0",
+    "@grafana/plugin-ui": "0.16.1",
```
muradsater, line 130: Exact pin is the right call: `@grafana/plugin-ui@0.17.0` peers `@grafana/data` / `@grafana/ui` at `^13.0.0`, which would jump our runtime floor to Grafana 13.

`renovate.json` extends `:automergeMinor`, and for 0.x a 0.16.1 → 0.17.x bump is a minor. Please ignore/pin this package in Renovate (or a one-line comment here) so that bump cannot automerge.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 This article documents the ongoing improvements we're making to the **Cognite
 Data Source for Grafana**.
 
+## Unreleased
+
+### Features
+
+- The minimum supported Grafana version is now 10.4.7, required by the shared editor components.
```
muradsater: This is a compatibility break, not a feature. Consider a **Breaking changes** (or similar) section: we are dropping Grafana 10.0.0–10.4.6. Also slightly premature as written — the shared editor components are not in this PR, so if this entry ships alone it oversells the reason for the floor bump.

yarn.lock
```diff
@@ -991,6 +991,11 @@
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.28.4.tgz#a70226016fabe25c5783b2f22d3e1c9bc5ca3326"
   integrity sha512-Q/N6JNWvIvPnLDvjlE1OUBLPQHH6l3CltCEsHIujp45zQUSSh8K+gHnaEX45yAT1nyngnINhvWtzN+Nb9D8RAQ==
 
+"@babel/runtime@^7.12.1", "@babel/runtime@^7.27.0", "@babel/runtime@^7.9.2":
+  version "7.29.7"
+  resolved "https://registry.npmjs.org/@babel/runtime/-/runtime-7.29.7.tgz#12022450c45a4da6d8d8287b18a4ff2ddb23f768"
```
muradsater: New lockfile entries (and a few rewritten existing ones: `classnames`, `lodash`, `@popperjs/core`, `nearley`) resolve to `registry.npmjs.org`. The rest of the file is `registry.yarnpkg.com`. Same packages in practice, but it contradicts the PR text and adds diff noise. Regenerating with the repo’s usual registry would keep this consistent.

jest.config.js
```diff
@@ -2,7 +2,18 @@
...
 module.exports = {
   // Jest configuration provided by Grafana scaffolding
   ...require('./.config/jest.config'),
+  // @grafana/plugin-ui ships an ESM-only nested uuid@14, and the GraphQL language
+  // service used for query autocomplete pulls in ESM-only vscode-languageserver-types.
+  transformIgnorePatterns: [
+    nodeModulesToTransform([
+      ...grafanaESModules,
+      '@grafana/plugin-ui',
+      'vscode-languageserver-types',
```
muradsater, line 16: Harmless to add now, but this is for the follow-up. `graphql-language-service` is already a dependency and current tests do not hit the ESM `vscode-languageserver-types` path. Fine as scaffolding; just flagging it is unused in this PR.

muradsater (commented):
Looks solid as a split PR: pin `@grafana/plugin-ui@0.16.1`, raise `grafanaDependency` to match its peers (`>=10.4.7`), and keep the yarn/jest workarounds in repo-owned files. CI is green through Grafana 13.2, and the e2e resolver already starts at 10.4.19.

A few things to land or acknowledge before this ships in a release — inline on the Grafana floor, the 0.16 pin vs Renovate, changelog wording, and lockfile registry mix. None of those are correctness/security blockers in the diff itself.

Nits not inlined:
- Observer stubs in `jest-setup.js` are the right place, but this PR’s unit suite does not open Grafana `Select` menus. Fine if the follow-up editor tests actually exercise them.
- PR body still says 10.0.13 “will be dropped”; this run already starts at 10.4.19.

src/plugin.json
evertoncolling: Agreed, nothing uses it yet. I put it back to `>=10.0.0` here. The bump will come with the editor PR, which has the first import.

On the major version: releases only go out on `v*` tags, so we can decide when that PR is ready to ship. I'd rather not go to 5.0.0 for this, but I'm open to it.

package.json
evertoncolling: Good catch. Added a rule in `renovate.json` that keeps it below 0.17.0, with the reason in the rule's description.

CHANGELOG.md
evertoncolling: Removed from this PR. It comes back in the editor PR under a "Breaking changes" heading and says which versions we drop (10.0.0 to 10.4.6).

yarn.lock
evertoncolling: Fixed. The 26 new entries now point to `registry.yarnpkg.com`, and the rewritten ones are back to what master has. The 8 `npmjs.org` entries still in the file were already on master. `yarn install --frozen-lockfile` passes.

jest.config.js
evertoncolling: Right, nothing here uses it yet. I checked on the editor branch: two test suites fail with `SyntaxError: Unexpected token 'export'` without this line. I'd like to keep it here so that PR only has editor code.

evertoncolling:
Thanks for the review. I replied inline and pushed three small commits. This PR now has no user-visible change.

On the two nits:
- Observer stubs: you're right that nothing here opens a `Select` menu. The variable editor tests later in the stack do, and they fail with `IntersectionObserver is not defined` without the stubs.
- PR description: updated, the 10.0.13 line was out of date.

muradsater (commented):
Re-reviewed the three follow-up commits. They address the previous comments:

- `grafanaDependency` stays at `>=10.0.0` until the editor PR (e2e is back on 10.0.13–13.2.2).
- Renovate cap `<0.17.0` is the right shape and the reason is in the rule.
- Changelog removed from this PR.
- New lockfile entries are on `registry.yarnpkg.com`.
- Keeping the `vscode-languageserver-types` Jest transform here is fine as scaffolding.

CI is green. This is ready from my side — no user-visible change, as intended.

src/plugin.json
muradsater: Looks good. Floor stays at 10.0.0 here; we can revisit a plugin major when the editor PR is ready to ship.

package.json
muradsater: This is the right control. `<0.17.0` still lets 0.16.x patches through.

CHANGELOG.md
muradsater: Agreed — better that it lands with the first import.

yarn.lock
muradsater: Looks consistent with the rest of the file now.

jest.config.js
muradsater: Fine to keep it here so the editor PR stays focused on editor code.

muradsater (approved):
Approve. Follow-ups addressed the review comments: no Grafana floor bump until the editor PR, Renovate cap below 0.17, changelog deferred, lockfile on registry.yarnpkg.com. CI is green.

PR #677, opened 2026-09-18
Add the GraphQL and instance-reference helpers behind the editors

evertoncolling:
Third of the five-PR stack (follows #675 and #676). Pure functions and datasource wiring only: no UI. The shared GraphQL editor (PR 4) and the variable editor (PR 5) are built on these, so this is the PR where the logic is worth reading closely; the two after it are mostly layout.

**Problem / User story**

The GraphQL panel and the GraphQL dashboard variable each carried their own copy of the same knowledge: the endpoint URL, how a result set is wrapped (`items` / `edges { node }`), how a dotted path is read off a row, how a `{{field}}` label token renders. The variable resolver lived as two private methods on `CogniteDatasource`, swallowed GraphQL errors into an empty list, and turned an object-valued field into the string `[object Object]`. A variable holding a data modelling instance could not be used in a GraphQL query at all: Grafana pastes the JSON reference verbatim, which is neither a GraphQL input object nor one of its parts.

**Solution**

New modules under `src/cdf/`, each a set of pure functions with its own tests:

- `instanceRef.ts` — the instance reference wire format (`encodeInstanceRef` / `parseInstanceRef`, moved out of `InstancePicker`, which re-exports them), path readers, and `readInstanceRef`, which reads a reference off a GraphQL row. A list, or a row with two nested objects that could both be the reference, yields nothing rather than an arbitrary pick.
- `graphqlInstanceRefs.ts` — `interpolateInstanceRefs` renders a variable holding references into GraphQL before Grafana's own interpolation: `${var}`, `$var` and `[[var]]` become an input object (a list for multi-value, **All** expanded from the variable's options), `${var.space}` / `${var.externalId}` the escaped part. Anything else, including tokens with a `:format`, is left for Grafana.
- `labelTokens.ts` — one `LABEL_TOKEN` regex, one `resolvePath` walker and `renderLabelToken`, now shared by the Time Series tab (`client.ts`) and `graphqlLabel.ts`, so a mistyped or empty token reads the same on both.
- `graphqlRows.ts` and `graphqlEndpoint.ts` — the row envelope and the endpoint path, stated once; the path segments are URL-encoded.
- `graphqlTimeSeries.ts` — `timeSeriesKeysFromSchema` finds the rows and row fields a data model's schema types as `CogniteTimeSeries` (or a type implementing it), by alias when one is used.
- `graphqlVariables.ts` — `runGraphqlQuery` and `metricFindGraphqlQuery`, extracted from the datasource. GraphQL errors now throw with their messages instead of resolving to nothing.

Wiring: `datasource.ts` interpolates instance references and the label for GraphQL panel queries and delegates `metricFindQuery` to the extracted resolver. `FlexibleDataModellingDatasource` reads rows through `graphqlRows`, names series through `interpolateGraphqlLabel` (falling back to `name`, which is what it used before), and plots an object the schema vouched for without requiring `type`.

**Dormant until later in the stack**

Flagging these up front, since nothing in this PR drives them yet:

- `FlexibleDataModellingQuery.schemaTsKeys` and `.label` are read by the datasource here and first written by the panel editor in PR 4. Unset, the datasource behaves exactly as on master.
- `VariableQueryData.displayField`, the `__instanceId` value field and `CogniteDatasource.runGraphqlQuery` are consumed by the variable editor and the preview pane in PRs 4 and 5.
- `interpolateInstanceRefs` is live for panel and variable queries, but the control that makes a variable hold references arrives in PR 5, so the feature is announced in that PR's changelog rather than here.

Following the #676 review, the changelog in this PR lists only what a user can observe once it merges.

**User-visible changes in this PR**

- A GraphQL variable whose query returns errors now shows them instead of resolving to an empty list. A dashboard with a quietly broken variable will start reporting it.
- A GraphQL variable whose value field is an object holding `space` and `externalId` resolves to the encoded reference instead of `[object Object]`.
- Time Series tab labels tolerate whitespace inside the braces (`{{ name }}`), as the GraphQL label does.

**Affected Resource types**

Data models via the GraphQL endpoint (panel queries and variable queries). Time Series tab label interpolation (shared token rules, no change for tokens without whitespace). No change to Events, Assets, Relationships or the legacy tabs.

Additive and optional, all backward compatible: `flexibleDataModellingQuery.schemaTsKeys?: string[]`, `flexibleDataModellingQuery.label?: string`, and `displayField?: string` on the variable query. Queries saved without them behave as before. `tsKeys` keeps its meaning (query-shape heuristics); schema-vouched keys are kept in the separate field so a legacy `__typename` key can never vouch for an untyped object.

`CHANGELOG.md` — the two bug fixes above. No user docs: nothing new is reachable from the UI yet.

**Have tests been updated?**

Yes, 79 new unit tests (suite is 499, up from 420): `instanceRef`, `graphqlInstanceRefs` (single, multi, All, custom all value, bracket syntax, escaping, scoped vars, tokens left for Grafana), `graphqlLabel`, `graphqlRows` (incl. endpoint encoding), `graphqlTimeSeries` (schema detection incl. aliases and `implements`, plotting rules, labels end to end through the datasource), and `metrics` for the variable resolver (references, object and list fields, display field, surfaced errors). The `@grafana/runtime` mock gains `getVariables()` and a word-boundary fix for `$name`.

Verified locally: frozen-lockfile install, typecheck, lint, unit suite, production build, and the config editor and feature flag e2e specs against Grafana 13.2. The data-panel e2e specs need the CI fixture project and are validated by CI.

**Screenshots and examples**

N/A — no UI in this PR.


muradsater (commented):
This is the right PR to read closely, and the extraction looks sound. Unit and compatibility checks passed; e2e was still running when I looked. It follows the #676 feedback: changelog only covers what a user can see once this merges, and `grafanaDependency` is left alone.

What looks right: JSON instance refs, all-or-nothing `readInstanceRef`, interpolating refs before Grafana's own pass, shared label-token rules (including whitespace), and keeping `schemaTsKeys` apart from `tsKeys`.

Three things to land or acknowledge — nested TS lists/connections (mostly a PR 4 concern), variable failures being broader than the changelog, and instance-ref interpolation being live already.

src/datasources/FlexibleDataModellingDatasource.ts
```diff
@@ -66,47 +71,64 @@ export class FlexibleDataModellingDatasource {
...
           });
-        }
-        // New detection for nested numeric time series
-        else if (
-          _.isObject(value) &&
-          (value as any).type === 'numeric' &&
-          (value as any).space != null &&
-          (value as any).externalId != null
-        ) {
+        } else if (isPlottable(value, schemaTsKeys.includes(key))) {
```
muradsater: Nested time series that arrive as a list or connection still will not plot. `_.isObject` is true for arrays, but an array (or a `{ items: [...] }` connection) has no `space`/`externalId`, so `isPlottable` skips it. `timeSeriesKeysFromSchema` also only vouches for row-level fields whose named type is `CogniteTimeSeries`, so `timeSeries { items { ... } }` is not keyed.

Harmless here because nothing writes `schemaTsKeys` yet. The panel editor in PR 4 will need to unwrap lists/connections or asset→timeSeries queries will still come back as tables.

src/cdf/graphqlVariables.ts
```diff
@@ -0,0 +1,132 @@
...
+      space: dataModel.space,
+      externalId: dataModel.externalId,
+      version: dataModel.version,
+    },
+  });
+
+  if (errors) {
+    // Surfaced, not just logged: a variable that silently resolves to nothing is
+    // indistinguishable from one whose query returned no rows.
+    throw new Error(`GraphQL variable query failed: ${formatGraphqlErrors(errors)}`);
```
muradsater, line 94: Surfacing GraphQL `errors` instead of `[]` is the right call, but this is broader than the changelog:

- A network/`fetchQuery` throw used to become `[]` and now fails the variable. Worth a changelog line.
- GraphQL can return partial `data` plus `errors`; this throws and drops the data.
- `formatGraphqlErrors` assumes an array. A non-array `errors` payload would throw `.map is not a function` instead of the GraphQL message.

src/datasource.ts
```diff
@@ -330,7 +335,14 @@ export default class CogniteDatasource extends DataSourceWithBackend<
         assetQuery?.target,
         eventQuery?.expr,
         templateQuery?.graphQlQuery,
-        flexibleDataModellingQuery?.graphQlQuery,
+        // Instance-reference variables become GraphQL syntax before Grafana
+        // pastes the remaining variables in verbatim.
+        flexibleDataModellingQuery?.graphQlQuery &&
+          interpolateInstanceRefs(
```
muradsater, line 341: Instance-ref interpolation is live on every Flexible Data Modelling panel query (and on GraphQL variable queries via `runGraphqlQuery`). The UI that *creates* ref-valued variables is PR 5, so the changelog defers the feature — but any existing variable whose value is already `{"space":"...","externalId":"..."}` will start being rewritten into GraphQL input-object syntax. Unlikely, but this is not fully dormant.

src/datasources/FlexibleDataModellingDatasource.ts
evertoncolling: Good catch, and it goes further: I introspected CogniteCore and every to-many relation is a `…DirectRelationConnection`, so the flat form does not even validate. Fixed here rather than in PR 4, since both files are in this PR: the schema lookup now vouches for a connection of time series, and the datasource reads every object in a connection or a plain list (a160ea5). The examples in PRs 4 and 5 used the flat form and are corrected too.

src/cdf/graphqlVariables.ts
evertoncolling: Added the request-failure case to the changelog and made `formatGraphqlErrors` tolerate a non-array payload, both with tests (a160ea5). I would like to keep throwing on partial data: the panel path treats any `errors` the same way, and a variable has nowhere to show a warning, so a partial option list would be a silently wrong variable.

src/datasource.ts
evertoncolling: True, it is live, and the PR description says so. I do not think it can change a working query though: a pasted JSON reference has quoted keys, so it was never valid GraphQL, bare or inside a string, and tokens with a `:format` are left alone. Only queries that were already broken get rewritten. Happy to gate it if you see a case I am missing.

muradsater (approved):
Re-reviewed a160ea5. The follow-up addresses the previous comments:

- Nested to-many relations: schema lookup vouches for a connection of time series, and the datasource plots every object in a connection or a plain list.
- Variable failures: changelog covers request failures as well as GraphQL errors; `formatGraphqlErrors` tolerates a non-array payload. Throwing on partial data is the right call for a variable.
- Live interpolation: a pasted JSON ref was never valid GraphQL, so gating is unnecessary.

CI is green through Grafana 13.2. Approve.

src/datasources/FlexibleDataModellingDatasource.ts
muradsater: Looks good. Connection unwrapping plus the schema vouch is the right fix, and doing it here rather than in PR 4 is correct.

src/cdf/graphqlVariables.ts
muradsater: Changelog and the non-array formatter look right. Agreed on throwing for partial data — a variable has nowhere to show a warning.

src/datasource.ts
muradsater: Agreed — no need to gate it. Quoted-key JSON was never valid GraphQL, and `:format` stays with Grafana.

